PR #9, opened 2021-10-31
Add Group Store

ramacatur:
Add group chat to store.


tulir:
I don't think this is necessary since the group list can just be requested from the server with `Client.GetJoinedGroups()` (the store is primarily meant for things that are not easily fetchable)

ramacatur:
> I don't think this is necessary since the group list can just be requested from the server with `Client.GetJoinedGroups()` (the store is primarily meant for things that are not easily fetchable)

Ok, thanks. I will try use GetJoinGroups

PR #25, opened 2021-11-17
data race issue and retry send message

Valdenirmezadri:
retry send when first message gets pair key error
canRetry - if account is a verified account, an retry looping issue occur

fix data race issue when check if isConnected

### DIFF
--- a/client.go
+++ b/client.go
@@ -59,8 +59,8 @@ type Client struct {
 	// even when re-syncing the whole state.
 	EmitAppStateEventsOnFullSync bool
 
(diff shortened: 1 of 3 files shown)


tulir (changes requested):
I'll probably look into retry receipts again in the near future, so I'll fix it properly myself and close this if you don't happen to do it sooner

tulir:
Retry receipts should work with https://github.com/tulir/whatsmeow/commit/7de34efe9e452ca3e3ae81c23737eac363e38e85 (and https://github.com/tulir/whatsmeow/commit/8752ed0761b5d82f21de4735601a53ba57a30cf3 makes isLoggedIn atomic)

PR #27, opened 2021-12-08
Update Protobuf Extract

adiwajshing:
1. Adds automatic duplicate spec removal
2. Removes the "Spec" suffix using a regex
3. Writes the proto to a file

### DIFF
--- a/binary/proto/extract/.gitignore
+++ b/binary/proto/extract/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+.DS_Store
+WAProto.proto
--- a/binary/proto/extract/index.js
+++ b/binary/proto/extract/index.js
@@ -1,6 +1,7 @@
 const request = require("request-promise-native");
 const acorn = require("acorn");
 const walk = require("acorn-walk");
+const fs = require('fs')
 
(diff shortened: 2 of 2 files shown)


tulir:
> Adds automatic duplicate spec removal

I don't think just ignoring duplicates works. There's 2 `Details` (renamed to [NoiseCertificateDetails](https://github.com/tulir/whatsmeow/blob/main/binary/proto/def.proto#L176-L183) and [VerifiedNameDetails](https://github.com/tulir/whatsmeow/blob/main/binary/proto/def.proto#L257-L264)) and 2 `MediaData` (renamed to [PBMediaData](https://github.com/tulir/whatsmeow/blob/main/binary/proto/def.proto#L1497) and [MediaData](https://github.com/tulir/whatsmeow/blob/main/binary/proto/def.proto#L1679-L1681)) which have conflicting names but different content. [`AppVersion`](https://github.com/tulir/whatsmeow/blob/main/binary/proto/def.proto#L1589-L1596) is actually the only one that has the same content twice

adiwajshing:
For the long term, yes -- I agree ignoring duplicates does not make sense. However, right now the dup protos aren't used anywhere either in the proto itself or libs

Issue #23, opened 2021-11-16
Program crash when client logout from linked devices. 


amirsheikh:
This is the error:
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x82a465]
go.mau.fi/whatsmeow.(*Client).sendProtocolMessageReceipt(0xc000194000, {0xc00023a080, 0xc000740160}, {0xa159df, 0x8})
        /root/go/pkg/mod/go.mau.fi/whatsmeow@v0.0.0-20211109215855-2e8804d7e690/message.go:370 +0x145
go.mau.fi/whatsmeow.(*Client).handleProtocolMessage(0xe, 0xc0003ac340, 0xc0003ac340)
        /root/go/pkg/mod/go.mau.fi/whatsmeow@v0.0.0-20211109215855-2e8804d7e690/message.go:319 +0x106
created by go.mau.fi/whatsmeow.(*Client).handleDecryptedMessage
        /root/go/pkg/mod/go.mau.fi/whatsmeow@v0.0.0-20211109215855-2e8804d7e690/message.go:343 +0x487
exit status 2

moxovil:
+1 I have the same thing. Arrived is there any solution for this problem?

Stasich:
+1

Stasich:
https://github.com/tulir/whatsmeow/pull/47 It's works for me, thx:)

PR #47, opened 2022-02-04
fix panic when client logout from his device

moxovil:
Fix issue #23


tulir (approved):
I'd guess there are other places it can panic too, but this is probably the main problem

PR #115, opened 2022-05-05
Handling WhatsApp Business Messages

hrizal:


tulir:
That XML parsing feels really hacky since `XMLString` is only meant for debugging. Would be nice to have custom unmarshaling directly from `*waBinary.Node` to arbitrary structs, but that's probably a lot of work :thinking:

Diden05:
@hrizal And why did they close the PR? Everything works!

hrizal:
Ok then

hrizal:
need adjusment ...

Diden05:
@hrizal The rest of the functions are interesting, working with the catalog (create / update / delete) the product, send the product and the collection.
Is a sequel planned?

PR #155, opened 2022-06-05
Itsme adding addgroup command

nlitsme:
I added two commands to mdtest: `addgroup` and `jsonadd`, but create a new whatsapp group, add members, and set the topic.
`jsonadd` reads the group's spec from a .json file, `addgroup` takes it from the commandline.

I extended  `CreateGroup` to report which participants need to be invited, so the `addgroup` command can send invites those user after creating the group.

Also I fixed a bug in the binary encoding of Nodes with an empty attribute value.


nlitsme:
I resolved the merge conflicts.  and added a 'help' command.

MightyKiller1 (approved):

tulir:
I think I added the group invite error handling at some point already.

mdtest is really only meant for testing things while developing whatsmeow rather than a fully featured CLI client, so probably not going to add more commands there

Issue #170, opened 2022-06-18
The autoReconnect method may be able to add the number of retries


sodaling:
https://github.com/tulir/whatsmeow/blob/5718022af87c34233b152be10cf670ee42100018/client.go#L332

At present, the method will continue to try to reconnect after a delay after receiving err. Can we let it retry a certain number of times and return directly. Instead of this endless retry?

PR #171, opened 2022-06-18
Optimize autoReconnect logic

sodaling:
#170 
I tried to make simple changes here without affecting the original logic. What do you think?


putrafajarh:
it is better to reset AutoReconnectTimes to 0, when connect successfully?

PR #232, opened 2022-08-19
Added revoke message from other participants

Duckduckgot:


Duckduckgot:
Screenshot_20220821-153217_WhatsApp

a5r0n:
maybe  it's better to Add new method as is a breaking change?
@tulir what do u think?

tulir:
Small breaking changes are fine, although the `me` boolean parameter is redundant.

It doesn't really need a method at all though, you can send revoke protocol messages manually.

PR #288, opened 2022-11-28
Remove local replace directive from go.mod

elias-orijtech:
Replace directives disable external `go run ...` support:

```
% go run github.com/tulir/whatsmeow/mdtest@latest
go: github.com/tulir/whatsmeow/mdtest@latest (in github.com/tulir/whatsmeow/mdtest@v0.0.0-20221126173344-e660988acdbc):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

A better way to develop multiple modules concurrently is with Go 1.18 workspaces[0].

[0] https://go.dev/blog/get-familiar-with-workspaces

### DIFF
--- a/mdtest/go.mod
+++ b/mdtest/go.mod
@@ -16,5 +16,3 @@ require (
 	golang.org/x/crypto v0.0.0-20221012134737-56aed061732a // indirect
 	rsc.io/qr v0.2.0 // indirect
 )
-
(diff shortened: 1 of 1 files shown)


tulir:
Go workspaces don't work very nicely and mdtest should always use the latest version of whatsmeow, so I'll keep it as is

PR #299, opened 2022-12-12
Add file logging to the waLog package (util/log.go)

### LINKED ISSUE #244: Implement file logging in the waLog package
Proposed:

- Parallel to  the `Stdout` method, implement `File` that returns a logger to send to a file.
- Proposed arguments:
  - Filename: string
  - Append: bool (when true, the logfile gets appended)
  - Reopen: bool (when true, the file is opened before outputting to it and closed after. That way an external logrotator can ensure that the file doesn't grow too much and that old versions are gzipped or similar).

Let me know if you agree. If yes I'll get to work on a pull request.

Update: FWIW I implemented this in https://github.com/KarelKubat/whatsmeow, if you deem it worthy for a merge, then let me know.

KarelKubat:
Noteworthy changes:

- `waLog.File()` returns a file logger using the arguments: 
  - `module` (string), just like the `Stdout()` logger
  - `filename` (string): output file
  - `reopen` (bool): should the logfile be reopened if an external script rotates it away
  - `append` (bool): overwrite or append new logs
- Literal strings like `"DEBUG"` are now constants
- The methods that a `Logger` interface must fulfill (`Errorf()`, `Debugf()` and friends) now return an error (which is non-nil if a logfile can't be reopened after disappearing)
- Common functionality between the loggers is now in helpers, like `shouldOutput()` (to prevent emitting when a log level is too low) or `timestamp()` (consistent time stamping in the `Stdout()` and `File()` loggers)
- The examples have been modified to show how logging to a file can be configured
- `util/log_test.go` is added.

Closes #244

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -27,7 +27,14 @@ func eventHandler(evt interface{}) {
 }
 
 func Example() {
-	dbLog := waLog.Stdout("Database", "DEBUG", true)
+	mainLog, err := waLog.File("", waLog.DebugLevel, "/tmp/whatsmeow.log", false, false)
+	if err != nil {
+		panic(err)
+	}
+	// Alternatively for logging to stdout:
+	//  mainLog := waLog.Stdout("", waLog.DebugLevel, true)
+
+	dbLog := mainLog.Sub("Database")
 	// Make sure you add appropriate DB connector imports, e.g. github.com/mattn/go-sqlite3 for SQLite
 	container, err := sqlstore.New("sqlite3", "file:examplestore.db?_foreign_keys=on", dbLog)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


tulir:
That's a lot more complicated than I was expecting 🤔

KarelKubat:
> That's a lot more complicated than I was expecting 🤔

Most code comes from ensuring that multiple loggers pointing to one file don't interweave their outputs, which is something `stdout` doesn't even know about. If there is something you would like me to clarify or simplify, let me know.

tulir:
I'll probably switch to zerolog in the near-ish future, so log output won't need to be built in anymore

PR #349, opened 2023-02-28
block and unblock user

Valdenirmezadri:


user.go
```diff
@@ -519,3 +519,46 @@ func (cli *Client) usync(ctx context.Context, jids []types.JID, mode, context st
...
 	}
 }
+
+func (cli *Client) UnblockUser(target types.JID) (Blockedlist []types.JID, err error) {
+	return cli.updateBlock(target, false)
+}
+
+func (cli *Client) BlockUser(target types.JID) (Blockedlist []types.JID, err error) {
+	return cli.updateBlock(target, true)
+}
```
tulir, line 529:
```suggestion
func (cli *Client) UnblockUser(target types.JID) ([]types.JID, error) {
	return cli.updateBlock(target, false)
}

func (cli *Client) BlockUser(target types.JID) ([]types.JID, error) {
	return cli.updateBlock(target, true)
}
```

That return variable name is wrong (shouldn't start with uppercase) and it's unnecessary anyway

user.go
```diff
@@ -519,3 +519,46 @@ func (cli *Client) usync(ctx context.Context, jids []types.JID, mode, context st
...
+		Attrs: waBinary.Attrs{"action": action, "jid": target},
+	}}
+	resp, err := cli.sendIQ(infoQuery{
+		Namespace: "blocklist",
+		Type:      iqSet,
+		To:        types.ServerJID,
+		Content:   content,
+	})
+	if err != nil {
+		return []types.JID{}, err
```
tulir, line 549:
```suggestion
		return nil, err
```

tulir (commented):
Should there also be a method to get the current blocklist? 🤔

Valdenirmezadri:
> Should there also be a method to get the current blocklist? thinking

should, but unfortunately I haven't found a way to request the list

Stein-Hakase:
> > Should there also be a method to get the current blocklist? thinking
> 
> should, but unfortunately I haven't found a way to request the list

Maybe you can just fake an unblock request ( by giving any phone number or try it empty ), if success the response should be always the current blocked list!

Issue #82, opened 2022-04-03
I got the following error when using MySQL


ratnokustiawan:
panic: failed to upgrade database: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'byte
a NOT NULL CHECK ( length(noise_key) = 32 ),
                identity_key bytea NOT NULL C' at line 6

ratnokustiawan:
MySQL Version '5.7.37'

tulir:
MySQL isn't supported. There's a PR for it at #48, but I'm not sure if it's possible to support simultaneously with Postgres, and Postgres takes priority.

PR #48, opened 2022-02-04
Compatibility with MySQL (MariaDB) storage

hoehermann:
Hi

My project [purple-whatsmeow](https://github.com/hoehermann/purple-gowhatsapp/tree/whatsmeow) gained some traction based on your awesome work. I am getting requests to support MySQL for the storage. I looked into it. The necessary changes do not seem too drastic. However, all experienced and mentally sane people I talked to urged me to use a DBMS. "Adjustments with regexes would be insane", they said. But here we are anyway. 

I tested my changes against SQLite and MySQL, but **not** with PostgreSQL. I would consider these changes experimental. [Table creation](https://github.com/tulir/whatsmeow/files/8002679/mysql_tables.sql.txt) and upgrades would have to be done manually, but for now "it is working for me"™.

What do you think?

Kind regards
Hermann

PR #363, opened 2023-03-21
Query adapter to easy compatibility with various drivers and implement Mysql adapter

caioreix:
This pull request adds a new layer to the queries which will enable easy compatibility with various drivers.

### DIFF
--- a/store/sqlstore/container.go
+++ b/store/sqlstore/container.go
@@ -15,6 +15,7 @@ import (
 
 	waProto "go.mau.fi/whatsmeow/binary/proto"
 	"go.mau.fi/whatsmeow/store"
+	"go.mau.fi/whatsmeow/store/sqlstore/query"
 	"go.mau.fi/whatsmeow/types"
 	"go.mau.fi/whatsmeow/util/keys"
 	waLog "go.mau.fi/whatsmeow/util/log"
@@ -25,6 +26,7 @@ type Container struct {
 	db      *sql.DB
 	dialect string
 	log     waLog.Logger
+	query   query.Adapter
 
 	DatabaseErrorHandler func(device *store.Device, action string, attemptIndex int, err error) (retry bool)
 }
@@ -34,6 +36,8 @@ var _ store.DeviceContainer = (*Container)(nil)
 // New connects to the given SQL database and wraps it in a Container.
 //
 // Only SQLite and Postgres are currently fully supported.
+//
+// You can support new drivers implementing query.Adapter and calling NewWithQueryAdapter.
 //
 // The logger can be nil and will default to a no-op logger.
 //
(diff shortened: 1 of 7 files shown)


ahmedRSA:
Amazing work, i wonder why tulir haven't merged this yet?

caioreix:
If this is related to MySQL query maintenance, I suggest creating a discussion and storing the implementation of the driver there. This would eliminate the need for you to update the MySQL queries, while also providing an easy way for the community to utilize them. I would be grateful if you could provide me with a feedback.

PurpShell:
#48 ???
This is a duplicate of that PR but the implementation is vastly different.

caioreix:
> #48 ??? This is a duplicate of that PR but the implementation is vastly different.

So in this case, it’s not a duplicate, the focus of this request is different. The main goal is to provide support for custom drivers, and MySQL is just an implementation example. I don't remember if I used Hoe's queries or implemented my own, but thanks to @hoehermann for the initiative. Unfortunately, none of us received a response.

hoehermann:
This PR is more invasive, but superior to #48 since it would allow using any database (provided an adapter is provided). In #82, tulir mentioned that "postgres has priority". Any database-related pull request probably needs to prove that it will not break compatibility with postgres ever – which I think will be hard.

PurpShell:
I think some community-maintained storage could be a viable approach for MySQL support. Going beyond sqlite3 and PostgresSQL, the two-most advanced and used in the world for the use-cases outlined by this app, would only make it a more complicated thing. Other forms of storage should probably be in their own go packages. If the interface makes it hard to implement from the outside, that must be discussed and improved on. If it is possible to make a third-party package for any other storage containers, then the best option would be to make such a third-party package that can be used with whatsmeow

caioreix:
Thats the point of this merge request, it opens up the possibility for someone to implement the query interface externally, without the need for this implementation to be in Whatsmeow. The removal of the MySQL part was also suggested above

PurpShell:
this pull request should be reopened then

mateusfmello:
I have two contradictory opinions on the subject.

1. It is indeed very interesting for the project that it be flexible without compromising or making maintenance difficult;

2. If the discussion is about supporting MySQL, it makes no sense, since developers who know how to use MySQL basically also know how to use PostgreSQL and SQLite. Yes, they have their particularities, but the difficulty of learning when you know one of the three is basically zero compared to learning another database, so is it worth adding and maintaining an extra layer in the project?

Still on item 2, what other databases would be interesting and make sense to use with WhatsMeow?

Are there many other options that make sense? Considering that WhatsMeow itself manages the database (focusing on data and data structure, not on infrastructure).

From the infrastructure perspective, providing a database is not as complex as it used to be, so is not supporting other databases a barrier for new developers to adopt WhatsMeow in their personal projects?

I have always used PostgreSQL and MySQL in all my projects. Years ago, when I implemented WhatsMeow in a project, I did it with SQLite. Since the examples used SQLite, there were some minor issues, but nothing that a simple Google search couldn't solve.

**Conclusion:**

In fact, flexibility in how to store data is very useful for any project. However, considering that the project already supports SQLite and PostgreSQL, I believe that if this modification is merged, it should only include these two databases (since they are the current WhatsMeow standards), leaving MySQL out so as not to increase the time needed to make corrections and improvements to the project (even if it is only in the queries).

MySQL can be a separate example of how to use the adapter, so the work already done on this improvement would not be lost and would make it easier for those who want to use WhatsMeow with MySQL, but it would keep WhatsMeow free from the responsibility of maintaining the link with another database.

We are talking about an open source project, where it is supported by developers who have their own private lives on a daily basis, so with this addition the project becomes more flexible, but without increasing its maintenance time.

---

What do you think about my "50 cents" on the subject?

hoehermann:
> If the discussion is about supporting MySQL, it makes no sense, since developers who know how to use MySQL basically also know how to use PostgreSQL and SQLite. Yes, they have their particularities, but the difficulty of learning when you know one of the three is basically zero compared to learning another database, so is it worth adding and maintaining an extra layer in the project?

The discussion _does_ make sense. Support for MySQL is something the _users_ requested. Please keep in mind that users typically are not developers. I am perfectly happy with SQLite. However, SQLite reportedly has issues with large amounts of concurrent connections. If whatsmeow was implemented in such a way it did not rely on an SQL dialect that is specific to PostgreSQL, then we would be fine. I tried that with #48. It has little impact on the code, but still was considered "too much". I know tulir has to draw the line somewhere and apparently, this is it.

tulir:
Providing a custom store (even a non-SQL one) is already supported, using the `sqlstore` package isn't required.

PurpShell:
> Providing a custom store (even a non-SQL one) is already supported, using the `sqlstore` package isn't required.

yup that's what i guessed earlier

mateusfmello:
> However, SQLite reportedly has issues with large amounts of concurrent connections.

In fact, SQLite has limitations, I managed to solve the problems caused by this configuration `_journal_mode=WAL`.

[Documentation](https://sqlite.org/wal.html)

Full example: `/db.sqlite?_journal_mode=WAL`

PR #370, opened 2023-04-09
implement writing app state - pin/mute/archive chat

### LINKED ISSUE #31: Implement writing app state
If we write one of these functions, we write the rest. App state write so important.

crazycodezombie:
Fixes #31

In this pull request I added the implementation of writing the app state.

This adds support for:
- Archive chat
- Pin chat
- Mute chat

* other types of states (e.g. delete) can easily be added by adding a PatchInfo definition and few function similar to the pin/mute/archive for the required action.

Archive chat is more complicated than mute/pin and requires a MessageRange parameter which complicates the implementation and need more research for implementing the correct message range. For now I just used the latest message in this parameter and it seems to work, but can be risky in the future.

In addition I added "upgrade5" function to the store, but with the implementation of storing the last message for each chat it will require re-login from the user (and i didn't want to let the user implement the message-range part). not sure what can be a better solution for that.

### DIFF
--- a/appstate.go
+++ b/appstate.go
@@ -301,3 +301,79 @@ func (cli *Client) requestAppStateKeys(ctx context.Context, rawKeyIDs [][]byte) 
 		cli.Log.Warnf("Failed to send app state key request: %v", err)
 	}
 }
+
+func (cli *Client) ArchiveChat(target types.JID, action bool) error {
+	patch, err := cli.appStateProc.NewArchivePatchInfo(target, action)
+	if err != nil {
+		return fmt.Errorf("failed to set archive to %v for chat %v: %w", action, target.String(), err)
+	}
+	if err := cli.UpdateChat(patch); err != nil {
+		return fmt.Errorf("failed to set archived to %v for chat %v: %w", action, target.String(), err)
+	}
+	return nil
+}
+
+func (cli *Client) MuteChat(target types.JID, action bool, muteTime int64) error {
+	patch := appstate.NewMutePatchInfo(target, action, &muteTime)
+	if err := cli.UpdateChat(patch); err != nil {
+		return fmt.Errorf("failed to set mute to %v for chat %v: %w", action, target.String(), err)
+	}
+	return nil
+}
+
+func (cli *Client) PinChat(target types.JID, action bool) error {
+	patch := appstate.NewPinPatchInfo(target, action)
+	if err := cli.UpdateChat(patch); err != nil {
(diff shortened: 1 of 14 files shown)


appstate.go
```diff
@@ -301,3 +301,79 @@ func (cli *Client) requestAppStateKeys(ctx context.Context, rawKeyIDs [][]byte)
...
+		return err
+	}
+
+	respCollection := resp.GetChildByTag("sync", "collection")
+	respCollectionAttr := respCollection.AttrGetter()
+	if t, ok := respCollectionAttr.GetString("type", false); ok && t == "error" {
+		return fmt.Errorf("unable to update chat, got error response from server")
+	}
+	
+	return cli.FetchAppState(patch.Type, false, false)
```
crazycodezombie, line 378: To me, It seems to be the easiest way to store the new state after writing the state successfully, but when looking on whatsapp-web, it seems to fetch the state only on archive request, I guess on mute/pin they just storing the calculated state from here. anyway fetching the state from the server does the work.

message.go
```diff
@@ -315,6 +315,11 @@ func (cli *Client) handleHistorySyncNotification(notif *waProto.HistorySyncNotif
 			go cli.handleHistoricalPushNames(historySync.GetPushnames())
 		} else if len(historySync.GetConversations()) > 0 {
 			go cli.storeHistoricalMessageSecrets(historySync.GetConversations())
+
+			if historySync.GetSyncType() == waProto.HistorySync_INITIAL_BOOTSTRAP ||
```
crazycodezombie, line 319: Here I was not sure, HistorySync_INITIAL_BOOTSTRAP & waProto.HistorySync_RECENT(chunk-order-1) is the only cases we should store the last message

1337-sys:
I tested it works great. @tulir doesn't want to process history messages, so we can give the last message id and archive it, just like in read chat. Please share me your IBAN dude. I waited a year for this moment. You are a perfect <3

appstate.go
tulir: Yeah probably fine to do that for now

message.go
tulir: I removed the message storage entirely since it's not necessary

tulir (approved):

Reyuki-san:
> but with the implementation of storing the last message for each chat it will require re-login from the user

@crazycodezombie i still confused why it will require re-login? could you explain more detail about the reason? i apologize for my poor knowledge

PR #377, opened 2023-04-21
Support new `ClearChatAction`

burstfreeze:
I am 99% confident that previously `ClearChatAction` either mapped to `DeleteChat` or to `DeleteMessageForMe` but it does not do that anymore on my phone. I found that it now triggers the literal `ClearChatAction` so I have added the required functionality for that.

You can test this PR by adding

```
	case *events.ClearChat:
		fmt.Println(evt)
		fmt.Println(evt.Action)
```

to the `handler` in the `mdtest/main.go`. In a group chat, send a few messages and then `Clear Chat`.

The output will contain the same messageID repeatedly but that's probably a server-side weirdness.

### DIFF
--- a/appstate.go
+++ b/appstate.go
@@ -154,6 +154,9 @@ func (cli *Client) dispatchAppState(mutation appstate.Mutation, fullSync bool, e
 		if cli.Store.Contacts != nil {
 			storeUpdateError = cli.Store.Contacts.PutContactName(jid, act.GetFirstName(), act.GetFullName())
 		}
+	case appstate.IndexClearChat:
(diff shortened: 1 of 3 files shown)


Manjit2003:
ClearChat deletes only the messages from the chat. DeleteChat deletes the whole chat along with messages.
ClearChat won't remove the chat from your WhatsApp screen. But DeleteChat will remove everything

Manjit2003:
Just to clear up the messages

Reyuki-san:
@Manjit2003 do you have example implementation of delete chat? i am little bit confused for this feature

PR #394, opened 2023-05-31
send reactions

Valdenirmezadri:


mamur-rezeki:
Reaction is already part of message protocol, doesn't need specific function to be added.

tulir:
`RevokeMessage` is already deprecated so others like that will definitely not be added. `BuildReaction` might make sense, especially if combined with a `BuildMessageKey` to share the code with `BuildRevoke`.

wildy283:
thx

send.go
```diff
@@ -247,6 +247,40 @@ func (cli *Client) BuildRevoke(chat, sender types.JID, id types.MessageID) *waPr
...
 }
 
+// ReactionMessage sends an emoji as a reaction to a message
+//
+// To react your own messages, pass your JID or an empty JID as the second parameter (sender).
+//
+// To react someone else's messages when you are in a group, pass the message sender's JID as the second parameter.
+//
+// To remove a reaction, pass an empty reaction.
+func (cli *Client) ReactionMessage(chat, sender types.JID, id types.MessageID, reaction string) (SendResponse, error) {
```
tulir, line 257: This method needs to be removed

send.go
```diff
@@ -247,6 +247,40 @@ func (cli *Client) BuildRevoke(chat, sender types.JID, id types.MessageID) *waPr
...
+	return cli.SendMessage(context.TODO(), chat, cli.buildReaction(chat, sender, id, reaction))
+}
+
+func (cli *Client) buildReaction(chat, sender types.JID, id types.MessageID, reaction string) *waProto.Message {
+	key := &waProto.MessageKey{
+		FromMe:    proto.Bool(true),
```
tulir, line 263: This method should be exported, and a new `BuildMessageKey` should be added to share the code with `BuildRevoke`

tulir (changes requested):

PR #395, opened 2023-06-02
store/sqlstore/container.go: add a Close() method that closes the DB,…

ghosthound:
store/sqlstore/container.go: add a Close() method that closes the DB, helpful for ensuring files are closed.


tulir (approved):
Hmm, I guess that's simple enough that it can be added.

(for more complicated things, you should use `NewWithDB` and manage the `*sql.DB` instance yourself)

PR #421, opened 2023-07-12
Adding approve/reject request join

Duckduckgot:


group.go
```diff
@@ -174,6 +176,59 @@ func (cli *Client) UpdateGroupParticipants(jid types.JID, participantChanges map
 	return resp, nil
 }
 
+// GroupRequestParticipantsList get the list of participants waiting to join the specified group in WhatsApp.
+func (cli *Client) GroupRequestParticipantsList(jid types.JID) []*types.JID {
```
tulir, line 180: This should return proper errors.

If you know which types of errors the IQ sending is expected to return, you should add human-readable errors to `errors.go`, then return `wrapIQError(humanReadableErr, err)`

For missing tags, return an `ElementMissingError`

group.go
```diff
@@ -174,6 +176,59 @@ func (cli *Client) UpdateGroupParticipants(jid types.JID, participantChanges map
...
+				fmt.Printf("Possibly failed to parse %s element in membership_approval_request node: %+v", child.Tag, childAG.Errors)
+			}
+			participants = append(participants, &part)
+		}
+	}
+	return participants
+}
+
+// UpdateGroupRequestParticipants can be used to accept or reject requests to join WhatsApp groups.
+func (cli *Client) UpdateGroupRequestParticipants(jid types.JID, participantChanges map[types.JID]ParticipantChange) (*waBinary.Node, error) {
```
tulir, line 209: Rename this to `UpdateGroupParticipantRequests` to match the other renaming

group.go
```diff
@@ -174,6 +176,59 @@ func (cli *Client) UpdateGroupParticipants(jid types.JID, participantChanges map
...
+	content := make([]waBinary.Node, len(participantChanges))
+	i := 0
+	for participantJID, change := range participantChanges {
+		content[i] = waBinary.Node{
+			Tag: string(change),
+			Content: []waBinary.Node{{
```
tulir, line 215: Someone reported that this was the wrong way for `UpdateGroupParticipants`, did you test this with multiple users?

Apparently the correct way is

```xml
<promote>
  <participant>...</participant>
  <participant>...</participant>
</promote>
```

instead of

```xml
<promote><participant>...</participant></promote>
<promote><participant>...</participant></promote>
```

If that is the case here too, I think the parameter for this function should be two lists of JIDs: one for approve and one for reject (I assume nothing other than approving/rejecting can be done here?)

tulir (changes requested):

group.go
```diff
@@ -170,10 +178,67 @@ func (cli *Client) UpdateGroupParticipants(jid types.JID, participantChanges map
...
+				return nil, &ElementMissingError{Tag: "jid", In: "membership_approval_request"}
+			}
+			participants = append(participants, &part)
+		}
+	}
+	return participants, nil
+}
+
+// UpdateGroupParticipantRequests can be used to accept or reject requests to join WhatsApp groups.
+func (cli *Client) UpdateGroupParticipantRequests(jid types.JID, participantChanges map[types.JID]ParticipantChange) (*waBinary.Node, error) {
```
a5r0n, line 213: The `membership_requests_action` as only accepting "approve" or "reject" changes, am i right?
if so, maybe we need to Add some validation around that?

or even create 2 separated functions, `ApproveGroupParticipantRequests` & `RejectGroupParticipantRequests`, so we can pass simple `[]JID` instead of map

PR #427, opened 2023-07-17
Fix typo that messes up node

burstfreeze:
Was comparing bytes like crazy until I noticed the typo in the tag


tulir (approved):
oof, does just that make it work? 😹

burstfreeze:
Not yet but close 😅

PR #430, opened 2023-07-19
fix: don't close client on QR timeout when pair-code is used

a5r0n:


a5r0n:
@tulir do you know if there is any expiration for the pair codes?

burstfreeze:
A timeout is most likely still required. My findings on the normal Web page:

- when you click the "Link via phone number" thing it opens a new WebSocket instead of keeping up the existing one
- the "old" socket receives a close command and the connection drops
- the "new" socket is valid for I think 210 seconds before it sends an `xmlstreamend` which causes the website to launch the next socket
- repeat forever

So right now the best thing to do with your PR would be to close the WebSocket connection, open a new one and keep closing it every 150 seconds if you want to play it safe.

a5r0n:
@burstfreeze 
when a new socket created, new QR events will automatically pushed, so there is no reason to open a new one and don't use the already open one.

@tulir any thoughts?

burstfreeze:
@a5r0n I'm not sure what you mean by that. Are you talking about whatsmeow or about the web page now?

a5r0n:
At least from the code pov, any new socket that not logged in, will automatically get QR code events.
I guess is same for the web

PR #460, opened 2023-09-08
Added support to block/unblock contacts and receive block/unblock events

emacielxp:
# Block/Unblock support
Added support to block/unblock contacts and receive block/unblock events

## Changes:

> connectionevents.go
- Sync at local cache the blocked list when successfully connect

> mdtest/main.go
- Included commands to test the block/unblock feature:
-- listblocked: get the blocked list from whatsapp
-- block: block contact
-- unblock: unblock contact

- Included contact blocked status changed event listener

> send.go
- Included validation at SendMessage function if the contact is blocked to not send message for a blocked contact

> notification.go
- Included function to handle the whatsapp block/unblock event changes

> types/events/events.go
- Created the contact blocked status changed event type definition

> user.go
-  Created necessary functions to allow cache, block and unblock contacts

> this feature was already tested and works fine at:
- iphone business
- iphone normal
- android business
- android normal

### DIFF
--- a/client.go
+++ b/client.go
@@ -105,6 +105,9 @@ type Client struct {
 	userDevicesCache           map[types.JID][]types.JID
 	userDevicesCacheLock       sync.Mutex
 
+	blockedContactsCache     map[types.JID]bool
+	blockedContactsCacheLock sync.Mutex
+
 	recentMessagesMap  map[recentMessageKey]*waProto.Message
 	recentMessagesList [recentMessagesSize]recentMessageKey
 	recentMessagesPtr  int
@@ -190,6 +193,8 @@ func NewClient(deviceStore *store.Device, log waLog.Logger) *Client {
 
 		groupParticipantsCache: make(map[types.JID][]types.JID),
 		userDevicesCache:       make(map[types.JID][]types.JID),
+
+		blockedContactsCache: make(map[types.JID]bool),
 
 		recentMessagesMap:      make(map[recentMessageKey]*waProto.Message, recentMessagesSize),
 		sessionRecreateHistory: make(map[types.JID]time.Time),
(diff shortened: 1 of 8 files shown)


mateusfmello:
Excellent modifications, I tested it and everything seems to be perfect.

I only wonder about the project standards.

In `mdtest` the commands `pin`, `mute`, `archive`, `presence` etc. are a single command varying the arguments, for "pinned" or "unpinned", "muted" or "unmuted" etc., I believe that the "block" command should follow this same pattern, thus saving code, simplifying the project and maintaining the standard that has been implemented.

I suggest the command would be like this:

`block <jid> <action>`

Where "action" is a boolean.

The code would look like this:
- Would delete the "unblock" command;

Block command:

```
// /mdtest/main.go
case "block":
	if len(args) < 2 {
		log.Errorf("Usage: block <jid> <action>")
		return
	}
	target, ok := parseJID(args[0])
	if !ok {
		return
	}
	action, err := strconv.ParseBool(args[1])
	if err != nil {
		log.Errorf("invalid second argument: %v", err)
		return
	}

	blockedList, err := cli.BuildBlockContact(target, action)
	if err != nil {
		log.Errorf("Error blocking contact: %v", err)
	}

	log.Infof("%+v", blockedList)
}
```

In "/user.go"
- Delete the `UnblockContact` and `BlockContact` methods

Create the `BuildBlockContact` method, which would be like this:

```
// BuildBlockCSe vontact updates the contact blocked status to false or true
func (cli *Client) BuildBlockContact(user types.JID, block bool) ([]types.JID, error) {
	blockedList, err := cli.setContactBlockedStatus(user, block)
	if err != nil {
		return nil, err
	}

	return blockedList, nil
}
```

@emacielxp  if you like this idea I can make a PR for you if you prefer it that way.

PR #471, opened 2023-10-02
feat: example of multiple sessions

mateusfmello:
In order to facilitate the initiation of new WhatsMeow users, I am adding another example that demonstrates the use of the project with multiple sessions.


apoorvcodes:
hey is there any way to use our own db something like mongodb for sessions ?

mateusfmello:
> hey is there any way to use our own db something like mongodb for sessions ?

As far as I know, it only supports PostgreSQL, I don't see any point in using a non-relational database in this project.

huzairuje:
> > hey is there any way to use our own db something like mongodb for sessions ?
> 
> As far as I know, it only supports PostgreSQL, I don't see any point in using a non-relational database in this project.

can we use slice or in-memory golang instead ? and implement an interface for it ?

mateusfmello:
> can we use slice or in-memory golang instead ? and implement an interface for it ?

I believe it is possible, @tulir  can help you better with this issue, as I have never implemented anything like this with Whatsmeow.

huzairuje:
> > can we use slice or in-memory golang instead ? and implement an interface for it ?
> 
> I believe it is possible, @tulir can help you better with this issue, as I have never implemented anything like this with Whatsmeow.

yes i haseen the code on when do pair success it will call container.Save()  and the function Save() is only implement the sqlstore method, i think we can implement another interface on memory slice type.

mateusfmello:
> > > can we use slice or in-memory golang instead ? and implement an interface for it ?
> > 
> > 
> > I believe it is possible, @tulir can help you better with this issue, as I have never implemented anything like this with Whatsmeow.
> 
> yes i haseen the code on when do pair success it will call container.Save() and the function Save() is only implement the sqlstore method, i think we can implement another interface on memory slice type.

Do you know that you can use PostgreSQL and SQLite in memory?
If your intention is to have the persistence layer in memory you can use one of the two in memory, I believe it is the fastest and simplest way to achieve this goal.

Manjit2003:
Simply pass `:memory:` in the sqlite driver filename

lairhas:
Hello @mateusfmello, thank you. I've implemented your method and successfully managed to run multiple devices using a single binary.

Is there a method to determine which device an event is originating from?

mateusfmello:
> Hello @mateusfmello, thank you. I've applied your method and managed to run a variable number of devices using a single binary.
> 
> Is there a method to determine which device an event is originating from?

From what I remember, within the messages comes the JID (WhatsApp ID) of the device, if you don't have it when making the connection and informing the handlers, pass the JID as a parameter to the handler methods that need to know which device the event came from / message.

But I believe that the messages / events have the JID of the recipient / device, I'm just not sure at the moment.

lairhas:
> > Hello @mateusfmello, thank you. I've applied your method and managed to run a variable number of devices using a single binary.
> > Is there a method to determine which device an event is originating from?
> 
> From what I remember, within the messages comes the JID (WhatsApp ID) of the device, if you don't have it when making the connection and informing the handlers, pass the JID as a parameter to the handler methods that need to know which device the event came from / message.
> 
> But I believe that the messages / events have the JID of the recipient / device, I'm just not sure at the moment.

I have read more and found this on client.go:
```
// If you want to access the Client instance inside the event handler, the recommended way is to
// wrap the whole handler in another struct:

	type MyClient struct {
		WAClient *whatsmeow.Client
		eventHandlerID uint32
	}

	func (mycli *MyClient) register() {
		mycli.eventHandlerID = mycli.WAClient.AddEventHandler(mycli.myEventHandler)
	}

	func (mycli *MyClient) myEventHandler(evt interface{}) {
		// Handle event and access mycli.WAClient
        }
```

I modified the wrapper a little for my needs and solved my multi sessions problem with it.

hrizal:
Cool, i will try too

Pada Rab, 8 Mei 2024 15.36, lairhas ***@***.***> menulis:

PurpShell:
@tulir close

PurpShell (changes requested):
I think this change complicates the structure of this file, and introduces concepts like this "fake database" that are not realistic onto how a person would usually configure whatsmeow.. 

This PR will serve to obscure how whatsmeow actually works, the main way people interact with whatsmeow now is via the client_test and main example due to the deletion of mdtest.

The only thing people need to know to actually make multiple clients is the methods of the Container (https://godocs.io/go.mau.fi/whatsmeow/store/sqlstore#Container). 

We already have the note regarding that:

https://github.com/tulir/whatsmeow/blob/69ba055bef0f46be2c4684c6db7fd82921db14af/client_test.go#L39

Might as well add a basic Go programming "hello world" tutorial in this file at this point...

.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
PurpShell, line 1:
```suggestion
```

Also, you should only add this in your global Git config, instead of pushing a useless gitignore to every repository you touch: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

mateusfmello:
> I think this change complicates the structure of this file, and introduces concepts like this "fake database" that are not realistic onto how a person would usually configure whatsmeow..
> 
> This PR will serve to obscure how whatsmeow actually works, the main way people interact with whatsmeow now is via the client_test and main example due to the deletion of mdtest.
> 
> The only thing people need to know to actually make multiple clients is the methods of the Container (https://godocs.io/go.mau.fi/whatsmeow/store/sqlstore#Container).
> 
> We already have the note regarding that:
> 
> https://github.com/tulir/whatsmeow/blob/69ba055bef0f46be2c4684c6db7fd82921db14af/client_test.go#L39
> 
> Might as well add a basic Go programming "hello world" tutorial in this file at this point...

Honestly, I don't see any complexity in reading the changes, but this was the way I found to exemplify the use of multiple sessions and a single session, but if you have a better way in mind, it will be very useful for the project.

"selectJIDFromFakeDatabase" was the way I used to exemplify the use of an identifier to retrieve the JID and proceed to connect the client, it wouldn't need to be a database, and the parameter wouldn't need to be the JID itself, it could be a username, email, UUID, any value that the developer uses to identify their users.

But we can really improve this part of retrieving the JID and add notes to make it clear that this part is specific to each project and not to WhatsMeow.

Although there is [sqlstore#Container](https://godocs.io/go.mau.fi/whatsmeow/store/sqlstore#Container) and a note about new developers on the project not being able to understand how to use multiple sessions, generating doubts. There are people who need a coded example to understand and from my point of view it is more understandable this way.

What do you suggest for improvements?

PurpShell:
> > I think this change complicates the structure of this file, and introduces concepts like this "fake database" that are not realistic onto how a person would usually configure whatsmeow..
> > This PR will serve to obscure how whatsmeow actually works, the main way people interact with whatsmeow now is via the client_test and main example due to the deletion of mdtest.
> > The only thing people need to know to actually make multiple clients is the methods of the Container (https://godocs.io/go.mau.fi/whatsmeow/store/sqlstore#Container).
> > We already have the note regarding that:
> > https://github.com/tulir/whatsmeow/blob/69ba055bef0f46be2c4684c6db7fd82921db14af/client_test.go#L39
> > 
> > Might as well add a basic Go programming "hello world" tutorial in this file at this point...
> 
> Honestly, I don't see any complexity in reading the changes, but this was the way I found to exemplify the use of multiple sessions and a single session, but if you have a better way in mind, it will be very useful for the project.
> 
> "selectJIDFromFakeDatabase" was the way I used to exemplify the use of an identifier to retrieve the JID and proceed to connect the client, it wouldn't need to be a database, and the parameter wouldn't need to be the JID itself, it could be a username, email, UUID, any value that the developer uses to identify their users.
> 
> But we can really improve this part of retrieving the JID and add notes to make it clear that this part is specific to each project and not to WhatsMeow.
> 
> Although there is [sqlstore#Container](https://godocs.io/go.mau.fi/whatsmeow/store/sqlstore#Container) and a note about new developers on the project not being able to understand how to use multiple sessions, generating doubts. There are people who need a coded example to understand and from my point of view it is more understandable this way.
> 
> What do you suggest for improvements?

we could make this more intuitive in the library code itself, maybe like a listAllSessions thing and picking the first element of that instead of the whole "GetFirstDevice" thing

Issue #384, opened 2023-05-12
Large amount of memory when downloading media files


kaefpiton:
When I shoot the memory profile, I found a large consumption of resources for downloading media files:
![mem_profile](https://github.com/tulir/whatsmeow/assets/50020923/6ccb16e9-66ca-47e9-8e90-75f181983593)

ahmedRSA:
how did you make this image?

kaefpiton:
> how did you make this image?

go tool pprof
I found out that when downloading files, they are saved in RAM ...
close this issue

PR #490, opened 2023-11-16
Optimize downloadMedia

brunoferreiras:
I had the same problem as described on #384

This pull request introduces a significant optimization in a specific function of the code. I replaced the usage of io.ReadAll with io.Copy to improve memory efficiency.

In my tests, I used a document of 10Mb using the `go tool pprof` to get this.

Before this change:
image

After this change:
image

This change aims to optimize memory consumption when processing large amounts of data. Additionally, I included examples of using go tool pprof to demonstrate performance improvements.

### DIFF
--- a/download.go
+++ b/download.go
@@ -7,6 +7,7 @@
 package whatsmeow
 
 import (
+	"bytes"
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


tulir:
That doesn't make any sense. You need 10mb of memory to download a 10mb file into memory, so my first guess would be that the second pprof is hiding the actual memory usage somewhere else. `bytes.Buffer` also shouldn't be any different from `io.ReadAll`, both of them have a byte slice as a buffer and expand it dynamically as they read data.

A potential small optimization would be pre-allocating the correct size of byte array and using `io.ReadFull`, rather than having `bytes.Buffer` or `io.ReadAll` dynamically grow a byte slice.

PR #491, opened 2023-11-17
Add label support

arugaz:


ahmedRSA:
bravo!! is this stable enough to test?

arugaz:
> bravo!! is this stable enough to test?

yes, but I think it should be cached

ahmedRSA:
> > bravo!! is this stable enough to test?
> 
> yes, but I think it should be cached

hmm lets wait for tulir response what does he think about this PR

appstate.go
```diff
@@ -223,6 +223,41 @@ func (cli *Client) dispatchAppState(mutation appstate.Mutation, fullSync bool, e
 			Action:       mutation.Action.GetUserStatusMuteAction(),
 			FromFullSync: fullSync,
 		}
+	case appstate.IndexLabelEdit:
+		act := mutation.Action.GetLabelEditAction()
+		eventToDispatch = &events.LabelEdit{
```
tulir, line 228: Should there be a label `BuildLabelEdit` to match these incoming events?

tulir (approved):
Looks simple enough

appstate.go
arugaz: > Should there be a label `BuildLabelEdit` to match these incoming events?

yeah there it is, i'll try

arugaz:
should i add helper on types for label color id ?

it would be 
```go
package types

const (
	LabelColorRed             int32 = iota
        ...
)
```

tulir:
Are the colors some arbitrary constants WhatsApp invented, or something standard like rgb? If it's custom then there should be constants in whatsmeow, otherwise not (but might make sense to use the color package https://pkg.go.dev/image/color)

arugaz:
> Are the colors some arbitrary constants WhatsApp invented, or something standard like rgb? If it's custom then there should be constants in whatsmeow, otherwise not (but might make sense to use the color package https://pkg.go.dev/image/color)

the colors its just constants id, whatsapp has 20 various colors for labels `0...19`

tulir:
wtf, the same color IDs are different based on whether the main device is Android or iOS?

```js
ANDROID_LABEL_COLOR_PALETTE = ["#FF9485", "#64C4FF", "#FFD429", "#DFAEF0", "#99B6C1", "#55CCB3", "#FF9DFF", "#D3A91D", "#6D7CCE", "#D7E752", "#00D0E2", "#FFC5C7", "#93CEAC", "#F74848", "#00A0F2", "#83E422", "#FFAF04", "#B5EBFF", "#9BA6FF", "#9368CF"]
IPHONE_LABEL_COLOR_PALETTE = ["#A62C71", "#90A841", "#C1A03F", "#792138", "#AE8774", "#F0B330", "#B6B327", "#C69FCC", "#8B6990", "#FF8A8C", "#54C265", "#FF7B6B", "#26C4DC", "#57C9FF", "#74676A", "#7E90A3", "#5696FF", "#6E257E", "#7ACBA5", "#243640"]
```

Constants might not make sense if the same value is a different color depending on the main device

PR #510, opened 2023-12-06
Added Secret when sending a conversation and image messages in mdtest to allow reactions in Announcements

Jonny1987:
The Announcements group in a community doesn't allow reactions to messages unless a MessageSecret is also sent so this fixes that for the sending of conversation message and image message in mdtest.

### DIFF
--- a/mdtest/main.go
+++ b/mdtest/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/mdp/qrterminal/v3"
 	"google.golang.org/protobuf/proto"
 
+	"go.mau.fi/util/random"
 	"go.mau.fi/whatsmeow"
 	"go.mau.fi/whatsmeow/appstate"
 	waBinary "go.mau.fi/whatsmeow/binary"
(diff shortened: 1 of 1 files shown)


tulir:
mdtest is meant as a test script, you should just add the message secrets in your own program.

It might make sense to add it automatically inside `SendMessage`, but it should probably only be added for community announcement groups, which makes it a bit more complicated to add there.

Jonny1987:
I understand but I am actually using another project which is using mdtest directly: https://github.com/HunterXProgrammer/Tasker-MdtestV5

It may not be completely correct to use mdtest like this but this other project is working well (apart from this bug).

Please can you merge my simple change to fix this if I create the pull request again?

tulir:
That project seems to have some very cursed manual patching going on, so it probably breaks whenever mdtest is changed. It should just be updated to have its own copy of mdtest with whatever modifications necessary instead of doing extremely hacky patching like that

PR #511, opened 2023-12-07
feat(sqlstore): Enable foreign key contraint on sqlite db

ch3ck:
# Description

- [x] There exists edge cases where foreign key contrainst is not automatically enabled on a sqlite db. This provides an easy way to recover from that method. An alternative is to export the db object which is not what you want.
- [x] Enable dependabot for package
- [x] Bump old packages(security fixes)


tulir:
I don't see how this helps at all. The docs already say how to enable foreign keys https://pkg.go.dev/go.mau.fi/whatsmeow/store/sqlstore#New. What may help is enforcing that they were enabled. One of the database upgrades already does that, but it could be an unconditional check on startup rather than only in that one schema upgrade.

Also, enabling foreign keys is a per-connection thing, it's not saved anywhere (some other things, like WAL mode, get saved in the database on disk. Foreign keys being enabled is not saved). Go uses multiple "connections" even for SQLite, which means just executing the pragma won't work, it has to be done in the connection string or a connect hook. The former is already documented, the latter will become an option in the future when I switch whatsmeow to using go.mau.fi/util/dbutil (which has a `sqlite3-fk-wal` connect hook: https://github.com/mautrix/go-util/blob/main/dbutil/litestream/register.go#L35-L51)

Dependabot is far too spammy and not particularly useful, definitely not going to add for go modules. It might make sense for github actions though, since those are easier to forget about. The recent release of gorilla/websocket also seems somewhat buggy, not going to upgrade that before a few more releases.

ch3ck:
I see @tulir whats a good example for using a Postgres db instead of SQLite?

Also, why do we need to store the session in an encrypted db? Is there a way to use a mounted volume instead of?

Issue #518, opened 2023-12-13
App states regualr_low and regular_high not syncing after archiving chat


MiguelArgentina:
In our app we are trying to add the `archive chat` feature for every message we send.
The first time we archive the chat, it works, but after that, every other message sent fails archiving with the errors shown below.
This is a part of the function where the message is sent and archived:
```go
        userJID := types.NewJID(phoneNumber, types.DefaultUserServer)
	message := &waProto.Message{
		Conversation: proto.String(messageContent),
	}
        ctx := context.Background()
	resp, err := client.SendMessage(ctx, userJID, message)
	chatID := resp.ID
	messageTimeStamp := resp.Timestamp
	lastMessageKey := &waProto.MessageKey{
		RemoteJid: proto.String(userJID.String()),
		FromMe:    proto.Bool(true),
		Id:        proto.String(chatID),
	}
	archiveChat := true

        archivePatch := appstate.BuildArchive(userJID, *proto.Bool(archiveChat), messageTimeStamp, lastMessageKey)
        err = client.SendAppState(archivePatch)
```

And these are the errors shown in the logs when `client.SendAppState(archivePatch)` is executed:

```go
<?xml version="1.0"?>
<collection has_more_patches="true" name="regular_low" type="error">
  <error code="409" text="conflict"/>
  <patches>
    <patch><!-- 867 bytes --></patch>
    <patch><!-- 1175 bytes --></patch>
    <patch><!-- 1021 bytes --></patch>
    <patch><!-- 1313 bytes --></patch>
    <patch><!-- 1329 bytes --></patch>
    <patch><!-- 1021 bytes --></patch>
    <patch><!-- 1021 bytes --></patch>
    <patch><!-- 1021 bytes --></patch>
   ...
  </patches>
</collection>
``` 

I also tried resyncing the patches (before archiving) with:
```go
        for _, name := range appstate.AllPatchNames {
		err := client.FetchAppState(name, true, false)
		if err != nil {
			client.Log.Errorf("Failed to do initial fetch of app state %s: %v", name, err)
		}
	}
```

but I get these two errors:

```
[Client ERROR] Failed to do initial fetch of app state regular_high: failed to decode app state regular_high patches: failed to verify patch v897: mismatching LTHash
```
and
```
[Client ERROR] Failed to do initial fetch of app state regular_low: failed to decode app state regular_low patches: failed to verify patch v712: mismatching LTHash
```
Does anyone have a tip on what this could be?
[...]

PR #519, opened 2023-12-19
cancel the execution of the method validateSnapshotMAC

matias-carpintini:


tulir:
Disabling validation is not a solution, the bug should be fixed instead

matias-carpintini:
we're desperate. could you check this: #518?

brunoaduarte:
Any updates to this? @matias-carpintini @tulir

Issue #182, opened 2022-06-24
Failed upgrading to v2 database because using pgx


SultanKs4:
## Problem
I try using [pgx](https://github.com/jackc/pgx) instead [pq](https://github.com/lib/pq) for postgresql but when upgrading to v2 database error happen:

```bash
08:33:38.709 [Database INFO] Upgrading database to v1
08:33:38.754 [Database INFO] Upgrading database to v2
panic: failed to upgrade database: ERROR: function instr(text, unknown) does not exist (SQLSTATE 42883)
```

### Reproduce

```go
import (
        ...
	_ "github.com/jackc/pgx/v4/stdlib"
}

func main() {
    ...
    container, err := sqlstore.New("pgx", databaseUrl, dbLog)
    ...
}
```

## Cause

https://github.com/tulir/whatsmeow/blob/bfb10aff31b90ba5a379b0d3652c041a694b0a3d/store/sqlstore/upgrade.go#L242-L246

because the dialect is `pgx` not `postgres` then executed `fillSigKeySQLite`

## Solution

Add check dialect if using `pgx`

```go
	if container.dialect == "postgres" || container.dialect == "pgx" {
		_, err = tx.Exec(fillSigKeyPostgres)
	} else {
		_, err = tx.Exec(fillSigKeySQLite)
	}
```

will PR to fix this issue ASAP

PR #523, opened 2024-01-03
Add support cloudsql-postgres dialect during db upgrade

ch3ck:
The container fails to upgrade database for `cloudsql-postgres` dialect. Same error as #182


ch3ck:
@tulir can you quickly review this, I'll need to pull the latest commit to unblock my workflow. 👍🏿

tulir (approved):

PR #541, opened 2024-02-06
Add the method to modify group add member mode

Narglc:


Narglc:
@tulir 
Just added a method for group settings, and I've tested it, it works well.
Could you please merge this? Is there anything I need to change?

PR #577, opened 2024-05-02
Improving the way app.js script is searched

JustSoursop:
The last script failed to find app.js because app.js is not directly loaded in the main page.

```log
/~/Dev/whatsmeow/binary/proto/extract/index.js:22
    const appID = index.match(/src="\/app.([0-9a-z]{10,}).js"/)[1]
                                                               ^

TypeError: Cannot read properties of null (reading '1')
    at findAppModules (/~/Dev/whatsmeow/binary/proto/extract/index.js:22:64)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /~/Dev/whatsmeow/binary/proto/extract/index.js:74:29

Node.js v18.19.0
```

### DIFF
--- a/binary/proto/extract/index.js
+++ b/binary/proto/extract/index.js
@@ -18,9 +18,23 @@ async function findAppModules(mods) {
         }
     }
     const baseURL = "https://web.whatsapp.com"
-    const index = await request.get(baseURL, ua)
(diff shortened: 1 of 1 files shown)


tulir:
This won't help, the script needs to be redone completely since whatsapp switched to a different js packaging system

PR #592, opened 2024-05-23
Add missing format specifier for stream error

torrca:
A logging cal on stream error code 515 was missing the format specifier for the connection error. This commit adds the specifier.


torrca:
Cheers

PR #596, opened 2024-05-24
Add ConnectWithTimeout method to support having a timeout when connecting.

ALiwoto:
This pull request adds a new method called `ConnectWithTimeout` to support having a timeout set when connecting to whatsapp. This `context.Conext` is later on passed to fs and is used when connecting (previously, we have been using `context.Backgound()` for the fs).
Can also use `context.Background()` if we do not want any timeout.

### DIFF
--- a/client.go
+++ b/client.go
@@ -395,9 +395,16 @@ func (cli *Client) WaitForConnection(timeout time.Duration) bool {
 	return true
 }
 
+// Connect connects the client to the WhatsApp web websocket with the specified context.
+// After connection, it will either authenticate if there's data in the device store,
+// or emit a QREvent to set up a new link.
+func (cli *Client) Connect() error {
+	return cli.ConnectWithTimeout(context.Background())
+}
+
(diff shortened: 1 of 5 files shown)


blaggacao:
Can you say a word or two about the motivation and concrete failure mode which triggered this?

I'm seeing timeout related issues and wonder if this was indirectly made to address a functionally equal failure mode as the one I observe.

<details>

I'm not currently [using a timeout](https://github.com/blaggacao/whatsmeow-native/blob/0150511fcce1efa500ece8065e9a3e9555e2e7c5/main.go#L106) on the service, but when connecting to the whatsmeow service socket, I use a timeout of 15 seconds within the application. This seems to cause cascading issues some times.

</details>

huzairuje:
@blaggacao i think this PR should be implemented, i have an issue sometimes the server got internet issue and we don't have any problem regarding how to connect using timeout or context cancelled.

PR #615, opened 2024-06-27
feature: Msmsg support (Meta AI and other personalities)

PurpShell:
Fully featured Meta AI implementation.

For people looking at this for implementation, please look at the MsgMetaInfo and MsgBotInfo parts of MessageInfo, if you are rendering something from this data, the TargetID in the meta should help you know which message to quote (while rendering the meta AI response)


message.go
```diff
@@ -88,6 +89,17 @@ func (cli *Client) parseMessageSource(node *waBinary.Node, requireParticipant bo
 		} else {
 			source.Chat = from.ToNonAD()
 		}
+	} else if from.IsBot() {
+		source.Sender = from
+		meta := node.GetChildByTag("meta")
+		ag = meta.AttrGetter()
+		targetChatJid := ag.JID("target_chat_jid")
```
tulir, line 97: https://go.dev/wiki/CodeReviewComments#initialisms (applies to a lot of places)

msgsecret.go
```diff
@@ -115,6 +118,38 @@ func (cli *Client) encryptMsgSecret(chat, origSender types.JID, origMsgID types.
...
+	secretKey := hkdfutil.SHA256(cli.applyBotMessageHKDF(messageSecret), nil, hkdfSecret, 32)
+
+	return secretKey
+}
+
+func (cli *Client) decryptBotMessage(messageSecret []byte, msMsg messageEncryptedSecret, messageId types.MessageID, targetSenderJid types.JID, info *types.MessageInfo) ([]byte, error) {
+	// gcm decrypt key generation
+	newKey := cli.createIncomingBotMessageSecret(messageId, targetSenderJid, info.Sender, messageSecret)
+
+	additionalData := []byte(fmt.Sprintf("%s\x00%s", messageId, info.Sender))
```
tulir, line 144: I think this is equivalent and makes the `createIncomingBotMessageSecret` function redundant?

```suggestion
	newKey, additionalData := cli.generateMsgSecretKey("", info.Sender, messageId, targetSenderJid, applyBotMessageHKDF(messageSecret))
```

types/jid.go
```diff
@@ -107,6 +109,11 @@ func (jid JID) IsBroadcastList() bool {
 	return jid.Server == BroadcastServer && jid.User != StatusBroadcastJID.User
 }
 
+func (jid JID) IsBot() bool {
+	match, _ := regexp.MatchString(`^1313555\d{4}$|^131655500\d{2}$`, jid.User)
+	return jid.Server == DefaultUserServer && match && jid.Device == 0
+}
```
tulir, line 115:
```suggestion
var botUserRegex = regexp.MustCompile(`^1313555\d{4}$|^131655500\d{2}$`)

func (jid JID) IsBot() bool {
	return jid.Server == DefaultUserServer && botUserRegex.MatchString(jid.User) && jid.Device == 0
}
```

msgsecret.go
PurpShell: I'll test it

types/jid.go
PurpShell: Makes sense

message.go
PurpShell: so no camel case

PurpShell: I'll review my own code using Github here to see where this really applies

msgsecret.go
PurpShell: I had considered this before, waited till u see the whole picture before I asked about it

message.go
tulir: It's still camel case, but initialisms are special and must be in consistent case rather than being treated like normal words

mdtest/main.go
```diff
@@ -45,7 +45,7 @@ var cli *whatsmeow.Client
 var log waLog.Logger
 
 var logLevel = "INFO"
-var debugLogs = flag.Bool("debug", false, "Enable debug logs?")
+var debugLogs = flag.Bool("debug", true, "Enable debug logs?")
```
PurpShell, line 48:
```suggestion
var debugLogs = flag.Bool("debug", false, "Enable debug logs?")
```
not again...

msgsecret.go
```diff
@@ -115,6 +123,18 @@ func (cli *Client) encryptMsgSecret(chat, origSender types.JID, origMsgID types.
 	return ciphertext, iv, nil
 }
 
+func (cli *Client) decryptBotMessage(messageSecret []byte, msMsg messageEncryptedSecret, messageId types.MessageID, targetSenderJid types.JID, info *types.MessageInfo) ([]byte, error) {
```
tulir, line 126:
```suggestion
func (cli *Client) decryptBotMessage(messageSecret []byte, msMsg messageEncryptedSecret, messageID types.MessageID, targetSenderJID types.JID, info *types.MessageInfo) ([]byte, error) {
```

message.go
```diff
@@ -200,10 +249,47 @@ func (cli *Client) decryptMessages(info *types.MessageInfo, node *waBinary.Node)
...
+			// step 1: get message secret
+			targetSenderJID := info.MsgMetaInfo.TargetSender
+			if targetSenderJID.User == "" {
+				// if no targetSenderJID in <meta> this must be ourselves (one-one-one mode)
+				targetSenderJID = cli.getOwnID()
+			}
+
+			messageSecret, err := cli.Store.MsgSecrets.GetMessageSecret(info.Chat, targetSenderJID, info.MsgMetaInfo.TargetID)
+			if err != nil || messageSecret == nil {
+				cli.Log.Warnf("Error getting message secret for bot msg with id %s", node.Attrs["id"].(types.MessageID))
```
tulir, line 264: Attributes can only be `string`, not `MessageID` and anyway better to use AttrGetter for safety

```suggestion
				cli.Log.Warnf("Error getting message secret for bot msg with id %s", node.AttrGetter().String("id"))
```

message.go
```diff
@@ -88,6 +90,17 @@ func (cli *Client) parseMessageSource(node *waBinary.Node, requireParticipant bo
 		} else {
 			source.Chat = from.ToNonAD()
 		}
+	} else if from.IsBot() {
+		source.Sender = from
+		meta := node.GetChildByTag("meta")
+		ag = meta.AttrGetter()
+		targetChatJID := ag.OptionalJIDOrEmpty("target_chat_jid")
+		if targetChatJID.User != "" {
+			source.Chat = targetChatJID.ToNonAD()
```
tulir, line 99:
```suggestion
		targetChatJID := ag.OptionalJID("target_chat_jid")
		if targetChatJID != nil {
			source.Chat = targetChatJID.ToNonAD()
```

message.go
```diff
@@ -88,6 +90,17 @@ func (cli *Client) parseMessageSource(node *waBinary.Node, requireParticipant bo
...
+		} else {
+			source.Chat = from
+		}
+		source.IsFromMe = false
```
tulir, line 103: No need to set this explicitly, it's false by default

```suggestion
```

message.go
PurpShell: oh 100%, oversight on my part

send.go
```diff
@@ -198,6 +202,64 @@ func (cli *Client) SendMessage(ctx context.Context, to types.JID, message *waPro
...
+	}
+
+	if isBotMode {
+		// common code for both inline and not inline modes
+		message.MessageContextInfo = &waE2E.MessageContextInfo{
+			BotMetadata: &waE2E.BotMetadata{
+				PersonaID: &req.BotPersonaID,
+			},
+			MessageSecret: random.Bytes(32),
+		}
```
tulir, line 227: This is what I meant by allowing custom message secrets:

```suggestion
		if message.MessageContextInfo == nil {
			message.MessageContextInfo = &waE2E.MessageContextInfo{}
		}
		if message.MessageContextInfo.MessageSecret == nil {
			message.MessageContextInfo.MessageSecret = random.Bytes(32)
		}
```

Also, since the persona ID already has a field in the protobuf, I don't think it needs a special extra param. If the persona ID is required, then it can be set if missing the same way as MessageSecret

send.go
```diff
@@ -130,6 +130,10 @@ type SendResponse struct {
 type SendRequestExtra struct {
 	// The message ID to use when sending. If this is not provided, a random message ID will be generated
 	ID types.MessageID
+	// JID of the bot to be invoked (optional)
+	InlineBotJID types.JID
+	// Persona ID for the bot if you are accessing a bot (optional)
+	BotPersonaID string
```
tulir, line 136:
```suggestion
```

send.go
```diff
@@ -198,6 +202,64 @@ func (cli *Client) SendMessage(ctx context.Context, to types.JID, message *waPro
...
+				},
+			}
+
+			messagePlaintext, deviceSentMessagePlaintext, marshalErr := marshalMessage(req.InlineBotJID, botMessage)
+			if marshalErr != nil {
+				err = marshalErr
+				return
+			}
+
+			participantNodes, _ := cli.encryptMessageForDevices(ctx, []types.JID{req.InlineBotJID}, ownID, resp.ID, messagePlaintext, deviceSentMessagePlaintext, waBinary.Attrs{})
```
tulir, line 258: The bot should never be the current user so no need to pass the device-sent wrapper through

```suggestion
			messagePlaintext, _, marshalErr := marshalMessage(req.InlineBotJID, botMessage)
			if marshalErr != nil {
				err = marshalErr
				return
			}

			participantNodes, _ := cli.encryptMessageForDevices(ctx, []types.JID{req.InlineBotJID}, ownID, resp.ID, messagePlaintext, nil, waBinary.Attrs{})
```

send.go
PurpShell: I agree with this (allow for custom message secret) in general but the persona ID seems to be required

send.go
PurpShell: alright, I was just keeping the standard

send.go
```diff
@@ -846,6 +913,11 @@ func (cli *Client) getMessageContent(baseNode waBinary.Node, message *waProto.Me
 			},
 		})
 	}
+
+	if botNode.Tag == "bot" && botNode.Content != nil {
```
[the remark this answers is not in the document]
PurpShell: ok

send.go
```diff
@@ -198,6 +200,70 @@ func (cli *Client) SendMessage(ctx context.Context, to types.JID, message *waPro
...
+	if !req.InlineBotJID.IsEmpty() {
+		if !req.InlineBotJID.IsBot() {
+			err = ErrInvalidInlineBotID
+			return
+		}
+		isInlineBotMode = true
+	}
+
+	isBotMode := isInlineBotMode || to.IsBot()
+	var botNode = waBinary.Node{
```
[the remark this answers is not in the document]
PurpShell: alright

PR #630, opened 2024-07-27
Fixed an issue when running mdtest as a systemd service since systemd services typically do not have a connected stdin.

cassioseffrin:
Here is my upgrade to allow the mdtest work as a linux systemctl service, the mainly modifications are here:

```
	// Check if stdin is available
	if isStdinAvailable() {
		handleStdin()
	} else {
		waitForInterrupt()
	}
```

Basically it will check for the presence of stdin and act accordingly. Below is a modified version of your main.go that should work better when running as a systemd service.

This fix will make this example work as a linux systemctl service:

```
[Unit]
Description=mdtest

[Service]
Type=simple
Restart=on-failure
RestartSec=230s
WorkingDirectory=/home/mdtest
ExecStart=/usr/local/go/bin/go run .
User=root
Group=root

[Install]
WantedBy=multi-user.target
```

### DIFF
--- a/mdtest/main.go
+++ b/mdtest/main.go
@@ -124,6 +124,23 @@ func main() {
 		return
 	}
 
+	// Check if stdin is available
+	if isStdinAvailable() {
+		handleStdin()
+	} else {
+		waitForInterrupt()
+	}
+}
+
+func isStdinAvailable() bool {
+	stat, err := os.Stdin.Stat()
+	if err != nil {
+		return false
+	}
+	return (stat.Mode() & os.ModeCharDevice) == 0
+}
+
+func handleStdin() {
 	c := make(chan os.Signal, 1)
 	input := make(chan string)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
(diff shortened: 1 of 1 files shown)


tulir:
mdtest is for testing, there's no reason to run it as a systemd service

cassioseffrin:
@tulir, I understand that mdtest is primarily intended for testing, but I believe that even test tools should be robust and versatile. The current implementation has a limitation when run in environments where stdin isn't available, such as when used as a systemd service. This is a common scenario in automated testing or continuous integration pipelines where services are managed by systemd.

The modification I’ve proposed addresses this limitation by ensuring that the program can gracefully handle the absence of stdin. This enhancement not only improves the flexibility of mdtest but also makes it more reliable in various testing environments. this PR would not only resolve a potential bug but also make the example more useful to other developers who might encounter similar situations. It could serve as a reference for how to handle stdin in Go programs that are intended to be run as system services

I have tested this changes in production, it's  batle tested for 3 weeks already and it's stable. It's a small modification in your example. Please considerer accept!

tulir:
I've deleted mdtest as people clearly keep misusing it for things it was never meant to do. If someone is maintaining high-quality wrappers around whatsmeow (CLI/HTTP/any other kind), I could consider listing them in the readme.

cassioseffrin:
@tulir Ok understood. Congrats for the great work on the meow project. People are difficult to deal. I am using your lib for a while, so far it's the best on github!

PR #632, opened 2024-08-01
Allow send command to send multi-line message separated by \n

RafayGhafoor:


mateusfmello:
I don't think this implementation is valid, `mdtest` is just for demonstration of usage, WhatsMeow already accepts the line break character, there is no need to use `\n`.

PR #702, opened 2024-11-20
appstate: add delete for me

marcos-travasso:


senpai295 (changes requested):
I think this makes more sense.

aryanchoudharypro:
what's the status on this pr?

PR #706, opened 2024-11-27
Add `Participants` to `MessageInfo`

Hellysonrp:
Messages sent to a broadcast list from the app have a list of participants, listing which contacts the message was sent to.

I need this information to correctly sync sent messages.

I don't know if I put this info in the right place, but this implementation is enough to satisfy my use case.
Feel free to modify if needed.


pjvds:
How do you currently get the participants of the broadcast list when sending to it?

Hellysonrp:
Superseded by f0915aa68985496f8b4670da71d05a06dd997fc2

PR #712, opened 2024-12-07
add MessageID on NewsletterMessage

RomySaputraSihananda:
## Commit Description:

**Add `MessageID` to `NewsletterMessage` struct**

This commit introduces the addition of the `MessageID` field to the `NewsletterMessage` struct. Previously, the struct only contained `MessageServerID`, `ViewsCount`, and `ReactionCounts`, without a unique identifier for the message itself. By adding the `MessageID`, each instance of `NewsletterMessage` can now store a unique message ID for better referencing.

### Key changes:
- Added a `MessageID` field of type `MessageID` to the `NewsletterMessage` struct.
- Updated the initialization of the `NewsletterMessage` struct in the relevant code, now retrieving the `MessageID` value from the `"id"` attribute in the source data.

With this change, it's now easier to identify and manage relevant messages, both for data retrieval and live message updates.

### DIFF
--- a/notification.go
+++ b/notification.go
@@ -274,6 +274,7 @@ func (cli *Client) parseNewsletterMessages(node *waBinary.Node) []*types.Newslet
 		}
 		msg := types.NewsletterMessage{
 			MessageServerID: child.AttrGetter().Int("server_id"),
+			MessageID:       child.AttrGetter().String("id"),
 			ViewsCount:      0,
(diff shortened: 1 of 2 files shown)


mateusfmello (approved):
all very well

RomySaputraSihananda (commented):
Add MessageID to NewsletterMessage struct

PR #721, opened 2024-12-26
optimize Send Message to Groups by changing Client => encryptMessageForDevices

Mahdi-Ostadrahim:
increased send message performance by collecting all neccessary device sessions and keys related to the message encryption and storing them into a cache, then storing them all into the db at once to minimize I/O to the database

### DIFF
--- a/send.go
+++ b/send.go
@@ -1014,10 +1014,21 @@ func (cli *Client) makeDeviceIdentityNode() waBinary.Node {
 	}
 }
 
+// Reducing the number of I/Os to the DB, improving performance by len(allDevices)*time(I/O) which is heavily noticable on large groups
+// one I/O to the db is 100ms on average, so sending in a group of 1000 people with at least 2 devices takes around 3 minutes to encrypt.
+// now it should take less than 4 seconds on sqlite, Less than 1 second on MSSQL and POSTGRES
 func (cli *Client) encryptMessageForDevices(ctx context.Context, allDevices []types.JID, ownID types.JID, id string, msgPlaintext, dsmPlaintext []byte, encAttrs waBinary.Attrs) ([]waBinary.Node, bool) {
 	includeIdentity := false
 	participantNodes := make([]waBinary.Node, 0, len(allDevices))
 	var retryDevices []types.JID
+	//Cache all sessions and identity keys relative to this message to decrease query time
+	//This will reduce the number of queries to the db from 3*len(allDevices) -> 2 queries only, heavily improving performance
+	var addresses []string
+	for _, jid := range allDevices {
+		addresses = append(addresses, jid.SignalAddress().String())
+	}
+	cli.Store.SessionsCache = cli.Store.Cache.GetSessions(addresses)
+	cli.Store.IdentityKeysCache = cli.Store.Cache.GetIdentityKeys(addresses)
 	for _, jid := range allDevices {
 		plaintext := msgPlaintext
 		if jid.User == ownID.User && dsmPlaintext != nil {
@@ -1041,6 +1052,13 @@ func (cli *Client) encryptMessageForDevices(ctx context.Context, allDevices []ty
 		}
 	}
 	if len(retryDevices) > 0 {
+		//By Making the commands to the db be done in a separate go routine,
+		//A bug will appear when none of the devices in "allDevices" have any sessions in the db,
+		//causing the program to use the old approach to get the session and not finding any
+		//causing the program to crash
+		//by filling the cashe with at least one element, the prekeys collected in bundles will also be stored in cache, avoiding this issue
+		cli.Store.SessionsCache[ownID.SignalAddress().String()] = []byte{}
+		cli.Store.IdentityKeysCache[ownID.SignalAddress().String()] = [32]byte{}
 		bundles, err := cli.fetchPreKeys(ctx, retryDevices)
 		if err != nil {
 			cli.Log.Warnf("Failed to fetch prekeys for %v to retry encryption: %v", retryDevices, err)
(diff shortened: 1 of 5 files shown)


HoshenKadosh:
@tulir can this be merged please?

store/store.go
```diff
@@ -126,6 +126,13 @@ type PrivacyTokenStore interface {
 	GetPrivacyToken(user types.JID) (*PrivacyToken, error)
 }
 
+type CacheStore interface {
+	GetSessions(addresses []string) map[string][]byte
```
devlikepro, line 130: Instead of using primitives (`string/byte`) we should try using already defined types, like `types.JID` (likely there's something for Session/IdentityKeys as well)

store/store.go
```diff
@@ -154,9 +161,14 @@ type Device struct {
...
 	MsgSecrets    MsgSecretStore
 	PrivacyTokens PrivacyTokenStore
+	Cache         CacheStore
 	Container     DeviceContainer
 
 	DatabaseErrorHandler func(device *Device, action string, attemptIndex int, err error) (retry bool)
+
+	//Cache to Temporary save sessions and identity keys for faster group send
+	SessionsCache     map[string][]byte
+	IdentityKeysCache map[string][32]byte
```
devlikepro, line 171: Looks like too many Caches here, I lost the idea when we use SessionCache, when we use IdentityKeysCache and when just Cache.

send.go
```diff
@@ -1066,7 +1084,21 @@ func (cli *Client) encryptMessageForDevices(ctx context.Context, allDevices []ty
...
+	//Store All Sessions at once. This decreases the number of commands to the database from 2*len(allDevices) -> 2 commands only
+	//An alternate, faster method would be using "go StoreSessions" which is incompatible with file databases such as sqlite
+	if len(cli.Store.SessionsCache) > 0 {
+		cli.Store.Cache.StoreSessions(cli.Store.SessionsCache)
+	}
+	if len(cli.Store.IdentityKeysCache) > 0 {
+		cli.Store.Cache.StoreIdentityKeys(cli.Store.IdentityKeysCache)
 	}
+	//clear the cache once the encryption is done to release memory
+	clear(cli.Store.SessionsCache)
```
devlikepro, line 1100: interesting that we clear cache after every call - why does it need to be available global then?

devlikepro (commented):
The solution likely will solve the issue, but the price we'll have to pay later on maintenance with that code is too much, IMO.

If we could hide all "caches" inside the Store (so the caller `encryptMessageForDevices` doesn't care about caches and always get the final result - it'd be great! And reducing number of "caches" would also help, right now you move entities from one to another in a bit complex way.

:information_source: It's not an official maintainer opinion, I was just passing by the PR :)

purpshell:
I think that, like @devlikepro said, having multiple ambiguous caches isn't very sustainable.

I'd suggest you implement your own SQLStore layer that hits your own custom cache first. You can also look into other options like Redis at that point.

Nubuki-all:
@tulir any chance you or any other maintainer are working on something that fixes this ?

tulir:
The proper solution is updating libsignal-protocol-go as well as the sql store to use contexts, after which the code can use a native db transaction. Not planned in the near future, but will probably be added at some point

store/store.go
Mahdi-Ostadrahim: IdentityKeys and Session types are both []byte in the code which is the reason I chose that

send.go
Mahdi-Ostadrahim: Since the I/O to the db is being used inside LoadSession and IsTrustedIdentity in store/signal.go, the check had to be made whether we already have it or not. Since both of these functions are whatsmeows implementation of another library, this was the only suitable place I could find to place it in

Mahdi-Ostadrahim:
> I think that, like @devlikepro said, having multiple ambiguous caches isn't very sustainable.
> 
> I'd suggest you implement your own SQLStore layer that hits your own custom cache first. You can also look into other options like Redis at that point.

Hello
Sorry for the late response
I didn't want to force the library to be dependent on a DB like redis
That said, I like the idea of merging the caches into a single cache entity. beyond that do you have any suggestions that improves this approach?

Mahdi-Ostadrahim:
> The proper solution is updating libsignal-protocol-go as well as the sql store to use contexts, after which the code can use a native db transaction. Not planned in the near future, but will probably be added at some point

If I'm not mistaken the approach you're suggesting the wrapping the entire process in a transaction, which wouldn't really solve the main issue which is too many I/Os to the database.
Or did I misinterpret something there?

Thank you for your response

purpshell:
> > I think that, like @devlikepro said, having multiple ambiguous caches isn't very sustainable.
> > I'd suggest you implement your own SQLStore layer that hits your own custom cache first. You can also look into other options like Redis at that point.
> 
> Hello Sorry for the late response I didn't want to force the library to be dependent on a DB like redis That said, I like the idea of merging the caches into a single cache entity. beyond that do you have any suggestions that improves this approach?

I meant implementing this on your own outside of the library, in my original approach.

theeusmartins:
i need this

GustavoGSA:
I try this PR in my fork and o have this error

image

GustavoGSA:
Individual mesage @http://s.whatsapp.net no working to with this PR, i send mutiples message and stop send

tay585 (approved):

Mahdi-Ostadrahim:
> I try this PR in my fork and o have this error
> 
> image

You have to share the version you're using for me to understand what's wrong in your fork
The line you have logged is 'queryValues += ","' in my fork which has no relation to the error you are displaying

GustavoGSA:
> > I try this PR in my fork and o have this error
> > image
> 
> You have to share the version you're using for me to understand what's wrong in your fork The line you have logged is 'queryValues += ","' in my fork which has no relation to the error you are displaying


I use the current version of the main branch, I forked it and then applied this PR

GustavoGSA:
I didn't understand, I forked the current version and merged this pull request and fixed all the conflicts, I didn't notice any difference, sending it to a group of 1000 people with the original version took 10s and with this PR it took 9.6s, it wasn't 1s as noted in the code.
I'm using PostgreSQL

PR #725, opened 2025-01-09
copy evt.Codes to avoid concurrency issues

devlikepro:
Right now in `emitQRs` function changes the original event, `events.QR`.
However, we send it already to other handlers, but we change it here.

In my case, I tried to `JSON.Marshal(evt)` and sometimes got `reflect: slice index out of range concurrency` error.

It looks like instead of changing `evt.Codes` we should copy it first and then change copied slice.

It's minor issue, but it's better not to change the values we already send outside, IMO.

Not sure if coping it before `Marshal` would help btw, because it'll be anyway concurrent access :thinking:

### DIFF
--- a/qrchan.go
+++ b/qrchan.go
@@ -59,9 +59,12 @@ type qrChannel struct {
 }
 
 func (qrc *qrChannel) emitQRs(evt *events.QR) {
+	// copy codes to avoid changing the original slice
+	codes := make([]string, len(evt.Codes))
+
 	var nextCode string
 	for {
-		if len(evt.Codes) == 0 {
+		if len(codes) == 0 {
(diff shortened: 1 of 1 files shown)


tulir:
It's probably easier to just not use the QR channel if you want to manage QR codes yourself 🤔

devlikepro:
@tulir the main idea - if you send event outside it should have "concurrent" access to it, not be changed after you sent it. 

You can close PR if you feel, fine for me!
We handled panic in JSON serialization, but I'd say it's good practice to avoid changing objects after sending it to consumers.

> if you want to manage QR codes yourself 

I don't, I love the way how qr channel does it.
I just wanted to dump all "events" from meow to JSON for troubleshooting/debug purpose.

tulir:
The other problem is that the PR is implemented incorrectly

PR #744, opened 2025-02-08
Added QR code output image to example client_test.go

cassioseffrin:
Description:
This PR refactors the existing Example function into a main function and introduces a new feature to save QR codes as images using the go-qrcode package. These changes improve usability by allowing users to generate and store QR codes for authentication.

Changes:
1. Added github.com/skip2/go-qrcode import for QR code generation.
2. Updated the QR code login event handling to include saving QR codes using saveQRCodeImage.
3. Implemented saveQRCodeImage function to store QR codes as qrcode.png.
4. Ensured proper error handling for QR code generation.

Benefits:
1. Users can now save QR codes for authentication instead of manually scanning them from the terminal.
2. The main function improves clarity and aligns with common Go practices.
3. Provides a cleaner separation of concerns by isolating QR code handling logic in saveQRCodeImage.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -13,6 +13,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/skip2/go-qrcode"
 	"go.mau.fi/whatsmeow"
 	"go.mau.fi/whatsmeow/store/sqlstore"
 	"go.mau.fi/whatsmeow/types/events"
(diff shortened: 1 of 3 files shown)


tulir:
This doesn't seem to do anything useful

hermanorigotto:
@tulir why this PR is not useful? I disagree with you !

Louise92rodriguez:
It seems like WhatsApp is detecting it and banning the numbers that uses this lib. I've already lost three SIM cards because of it. Better to get out this lib!

Diden05:
> It seems like WhatsApp is detecting it and banning the numbers that uses this lib. I've already lost three SIM cards because of it. Better to get out this lib!

If you can't use the product correctly, then maybe you shouldn't start?

PR #747, opened 2025-02-15
feat: add makefile and protobuf api updates

protomolequle:
# Description

- Upgrade generated go version files to latest version
- Add Makefile for local testing
- Remove unnecessary pb.raw files
- Add gitignore to avoid checking in unnecesary pb raw files, causes code base to bloat unnecessarily
- Depends on #746 getting merged first


tulir:
The makefile looks unnecessary, there's already pre-commit set up

The protobuf changes are incorrect, the pb.raw files are there intentionally, they replace the big byte arrays that are in the go code by default https://github.com/protomolequle/whatsmeow/blob/3f21ccdf15a1bb01ae2f690315ff73f102e872a3/proto/waAdv/WAAdv.pb.go#L422-L486

PR #748, opened 2025-02-15
feat: replace deprecated proto apis to latest api updates

protomolequle:
# Description

- This cleans up the codebase removing deprecated api references from code
-  Depends on #746 and #747 getting merged first


protomolequle:
@tulir excellent job with this pakage, we've been maintaining our own internal fork of this for the past few years and given how slow PRs are merged to upstream, I'd like to bring some of these changes upstream so everyone can benefit. But appreciate your dedication on this :-)

tulir:
The last commit looks useful, but it has a bunch of incorrect changes mixed in, so I'll pick out the correct changes manually

client_test.go
```diff
@@ -27,11 +27,8 @@ func eventHandler(evt interface{}) {
 }
 
 func Example() {
-	// |------------------------------------------------------------------------------------------------------|
-	// | NOTE: You must also import the appropriate DB connector, e.g. github.com/mattn/go-sqlite3 for SQLite |
-	// |------------------------------------------------------------------------------------------------------|
-
 	dbLog := waLog.Stdout("Database", "DEBUG", true)
+	// Make sure you add appropriate DB connector imports, e.g. github.com/mattn/go-sqlite3 for SQLite
```
tulir, line 31: The comment was intentionally big to make it harder to miss, so making it smaller isn't very helpful

go.mod
```diff
@@ -1,12 +1,10 @@
 module go.mau.fi/whatsmeow
 
-go 1.23.0
-
-toolchain go1.24.0
+go 1.24
```
tulir, line 3: Go 1.22 and up require the patch version in the `go` statement, so `go 1.24` is not valid. The `toolchain` statement is also needed so that the auto-update mechanism would get the latest version rather than the oldest supported version.

go.mod
```diff
@@ -1,12 +1,10 @@
...
 
 require (
 	github.com/google/uuid v1.6.0
-	github.com/gorilla/websocket v1.5.0
+	github.com/gorilla/websocket v1.5.3
```
tulir, line 7: The gorilla websocket package is not well maintained, they released a few bad versions, then a new version reverting most of the changes back. I'll just keep it on v1.5.0 until I get around to switching to https://github.com/coder/websocket

mediaretry.go
```diff
@@ -9,11 +9,12 @@ package whatsmeow
 import (
 	"fmt"
 
-	"go.mau.fi/util/random"
 	"google.golang.org/protobuf/proto"
 
+	"go.mau.fi/util/random"
+
```
tulir, line 15: `go.mau.fi/util` is an external module from whatsmeow's perspective, so it should be sorted with other external non-stdlib modules. Imports should always be sorted into stdlib, external (all other modules) and internal (current module)

message.go
```diff
@@ -304,7 +307,9 @@ func (cli *Client) decryptMessages(info *types.MessageInfo, node *waBinary.Node)
 			return
 		}
 		retryCount := ag.OptionalInt("count")
-		cli.cancelDelayedRequestFromPhone(info.ID)
+		if retryCount > 0 {
+			cli.cancelDelayedRequestFromPhone(info.ID)
+		}
```
tulir, line 312: `cancelDelayedRequestFromPhone` is a no-op if there's no pending request, so there's no need to check the retry count here

receipt.go
```diff
@@ -29,8 +29,9 @@ func (cli *Client) handleReceipt(node *waBinary.Node) {
 				}
 			}()
 		}
-		cli.dispatchEvent(receipt)
+		go cli.dispatchEvent(receipt)
 	}
+	go cli.sendAck(node)
```
tulir, line 34: `maybeDeferredAck` already handles sending acks. `dispatchEvent` is also intentionally synchronous to allow library users to determine whether they want to handle it in the background or not.

sendfb.go
```diff
@@ -34,61 +34,35 @@ import (
...
+	FBMessageApplicationVersion = 2
+	FBConsumerMessageVersion    = 1
+)
 
 // SendFBMessage sends the given v3 message to the given JID.
 func (cli *Client) SendFBMessage(
 	ctx context.Context,
 	to types.JID,
-	message armadillo.RealMessageApplicationSub,
+	message *waConsumerApplication.ConsumerApplication,
```
tulir, line 47: `SendFBMessage` is supposed to be able to send different types of messages, not only `ConsumerApplication` messages

store/clientpayload.go
```diff
@@ -60,24 +61,24 @@ func (vc WAVersionContainer) String() string {
 	return strings.Join(parts, ".")
 }
 
-// Hash returns the md5 hash of the String representation of this version.
-func (vc WAVersionContainer) Hash() [16]byte {
-	return md5.Sum([]byte(vc.String()))
+// Hash returns the SHA-256 hash of the String representation of this version.
+func (vc WAVersionContainer) Hash() [32]byte {
+	return sha256.Sum256([]byte(vc.String()))
```
tulir, line 66: The hash function can't just be arbitrarily changed, it has to match the one whatsapp uses

tulir (commented):
Specific problems in the last commit for reference (no action necessary, I already removed them from the merged commit)

sendfb.go
protomolequle: Makes sense

PR #749, opened 2025-02-17
Add BuildContact and CRUD functionality for contacts

fourjr:
This functionality does not work on all WhatsApp mobile apps. I suggest testing using a WhatsApp desktop client as the contact changes are reflected there.

Example Usage
```golang
// add or edit contact
client.SendAppState(appstate.BuildContact(v.Info.Sender, "Full Name"))
// remove contact
client.SendAppState(appstate.RemoveContact(v.Info.Sender))
```


AlejandroPerez92:
That is awesome! It could probably prevent some bans

appstate/encode.go
```diff
@@ -60,6 +63,27 @@ func BuildMute(target types.JID, mute bool, muteDuration time.Duration) PatchInf
 	}
 }
 
+func BuildContact(target types.JID, fullName string, add bool) PatchInfo {
```
AlejandroPerez92, line 66: What do you think about creating another method to remove the contact? It would be more self-explanatory and would take only the required arguments for that logic

fourjr: I added it this way as the other app state builders are done with this syntax. If the library prefers an alternate method, I wouldn't mind implementing as well.

AlejandroPerez92: I would suggest something like 
`func BuildContact(target types.JID, fullName string) PatchInfo`
`func RemoveContact(target types.JID) PatchInfo`

What do you think @tulir?

purpshell: +1

purpshell (changes requested):

appstate/encode.go
fourjr: My [latest commit](https://github.com/tulir/whatsmeow/pull/749/commits/36377340b2ac1e878f458fb201350c0761a251db) changes the syntax as per suggested.

purpshell (approved):
looks pretty good to me

appstate/encode.go
```diff
@@ -60,6 +63,37 @@ func BuildMute(target types.JID, mute bool, muteDuration time.Duration) PatchInf
 	}
 }
 
+func BuildContact(target types.JID, fullName string) PatchInfo {
```
devlikepro, line 66:
```suggestion
func BuildAddContact(target types.JID, fullName string) PatchInfo {
```

appstate/encode.go
```diff
@@ -60,6 +63,37 @@ func BuildMute(target types.JID, mute bool, muteDuration time.Duration) PatchInf
...
+				ContactAction: &waSyncAction.ContactAction{
+					FullName:                 &fullName,
+					SaveOnPrimaryAddressbook: proto.Bool(true),
+				},
+			},
+		}},
+	}
+}
+
+func RemoveContact(target types.JID) PatchInfo {
```
devlikepro, line 83:
```suggestion
func BuildRemoveContact(target types.JID) PatchInfo {
```

devlikepro (commented):
Idk how for others, but I'd expect method `RemoveContact` to actually "remove the contact".

For consistency with other methods, we could apply the changes, so the function name makes sense, because it actually "builds remove contact" data - so `BuildRemoveContact`

The same for `BuildAddContact`

Or even
- `BuildAddContactPatch`
- `BuildRemoveContactPatch`

a bit wordy, but show what the methods actually do

huzairuje:
nice to have and i am  abit lost when something with contacts management, LGTM

ashuvax:
any updates?

RakaiSeto:
this feature will be very good, looking forward to it

purpshell (approved):
This feature is stable now. For it to work though, you need to enable a little trigger on the main device. 

Note that that these contacts only stay on WhatsApp itself rather than becoming system-level, but it does make it better for keeping contacts separate if you have multiple numbers per device 👀.

Valdenirmezadri:
> This feature is stable now. For it to work though, you need to enable a little trigger on the main device.
> 
> Note that that these contacts only stay on WhatsApp itself rather than becoming system-level, but it does make it better for keeping contacts separate if you have multiple numbers per device 👀.

where is the little trigger?

purpshell:
> This feature is stable now. For it to work though, you need to enable a little trigger on the main device.
> 
> Note that that these contacts only stay on WhatsApp itself rather than becoming system-level, but it does make it better for keeping contacts separate if you have multiple numbers per device 👀.

Update on this, the feature is rolled out fully and is stable. What's delaying the merge of this PR?

guilhermejansen:
> > This feature is stable now. For it to work though, you need to enable a little trigger on the main device.
> > Note that that these contacts only stay on WhatsApp itself rather than becoming system-level, but it does make it better for keeping contacts separate if you have multiple numbers per device 👀.
> 
> Update on this, the feature is rolled out fully and is stable. What's delaying the merge of this PR?

It works great!

igortrinidad:
> This feature is stable now. For it to work though, you need to enable a little trigger on the main device. 
> 
> Note that that these contacts only stay on WhatsApp itself rather than becoming system-level, but it does make it better for keeping contacts separate if you have multiple numbers per device 👀.

What is the triagem please?

purpshell:
@fourjr Please resolve the merge conflicts

fourjr:
@purpshell merge conflicts resolved

igortrinidad:
Up

fedelombar:
why isn’t this getting merged?

PR #560, opened 2024-03-28
Call support

arugaz:

PR #753, opened 2025-02-21
Add OfferCall method (by @gamersindo1223)

fourjr:
Saw this in https://github.com/tulir/whatsmeow/pull/560 and thought that it might be useful to add in the main repo

### DIFF
--- a/call.go
+++ b/call.go
@@ -7,7 +7,14 @@
 package whatsmeow
 
 import (
+	"context"
+	"encoding/hex"
+	"fmt"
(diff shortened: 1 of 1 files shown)


tulir:
Doesn't seem particularly useful

fourjr:
I tried using that method and unfortunately `Client.encryptMessageForDevices` is not exposed, only `Client.encryptMessageForDevice`. Is there any way you can add this to the DangerousInternals @tulir?

PR #777, opened 2025-03-21
add event and prekey function

CharlesWone:
prekeys: query the last time the public key was uploaded
prekeys: publicize the FetchPreKeys function
events: add keep alive success event
events: logout event adds current jid


connectionevents.go
```diff
@@ -39,7 +39,7 @@ func (cli *Client) handleStreamError(node *waBinary.Node) {
 	case code == "401" && conflictType == "device_removed":
 		cli.expectDisconnect()
 		cli.Log.Infof("Got device removed stream error, sending LoggedOut event and deleting session")
-		go cli.dispatchEvent(&events.LoggedOut{OnConnect: false, Reason: events.ConnectFailureLoggedOut})
+		go cli.dispatchEvent(&events.LoggedOut{Jid: cli.Store.ID, OnConnect: false, Reason: events.ConnectFailureLoggedOut})
```
tulir, line 42: These are unnecessary and won't be added, the user is expected to remember which client an event came from

keepalive.go
```diff
@@ -54,6 +54,10 @@ func (cli *Client) keepAliveLoop(ctx context.Context) {
 					go cli.dispatchEvent(&events.KeepAliveRestored{})
 				}
 				lastSuccess = time.Now()
+				go cli.dispatchEvent(&events.KeepAliveSuccess{
+					Jid:         cli.Store.ID,
+					LastSuccess: lastSuccess,
+				})
```
tulir, line 60: What's the point of this? There are already events for keepalive failing and restoring

prekeys.go
```diff
@@ -131,6 +140,54 @@ func (cli *Client) fetchPreKeys(ctx context.Context, users []types.JID) (map[typ
 	return respData, nil
 }
 
+func (cli *Client) FetchPreKeys(ctx context.Context, users []types.JID) (map[types.JID]preKeyResp, error) {
+	return cli.fetchPreKeys(ctx, users)
+}
+
+func (cli *Client) FetchPreKeyTimes(ctx context.Context, users []types.JID) (map[types.JID]int64, error) {
```
tulir, line 147: And what's the point of this? The library handles prekeys internally, users don't need to care about them

tulir (changes requested):

PR #789, opened 2025-04-11
Add public access to the underlying database

rakibhoossain:
This PR adds a **GetDB()** method to store.Container, allowing external access to the underlying ***sql.DB**. This is useful for developers who need to perform advanced operations like custom queries, migrations.


devlikepro (commented):
I'd suggest using `NewWithDB` and saving the database reference somewhere else so you can use it later.

Why bothering exposing internals if you can "inject" the internal and do whatevery you need with this :thinking:

tulir:
Indeed, there's no reason to get it later when you can just make the db yourself. The change is wrong too as the CI says

PR #800, opened 2025-04-30
status: Add req.Participants for Status message

devlikepro:
Right now there's a problem when you try to send **status** (stories) message to contacts if you have > 50K contacts.
Web send such big messages in chunks, using the same message.id.

The RP add `req.Participants` we can send status message to many (>50K) contacts in chunks, using the same message.id

Note: May be we need to add some assertion about `req.Participants` being set and allow it only for `status@broadcast`, because right now it can confuse API a bit. Or may be name it like `req.StatusBroadcast[Receipents|Participants]` :thinking: 

Let me know how we can improve the changes, I'll do that!

### DIFF
--- a/send.go
+++ b/send.go
@@ -147,6 +147,9 @@ type SendRequestExtra struct {
 	MediaHandle string
 
 	Meta *types.MsgMetaInfo
+
+	// When sending status message you can specify the recipients
+	Participants []types.JID
 }
 
 // SendMessage sends the given message.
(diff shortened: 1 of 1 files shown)


send.go
```diff
@@ -304,11 +307,15 @@ func (cli *Client) SendMessage(ctx context.Context, to types.JID, message *waE2E
 				extraParams.addressingMode = types.AddressingModePN
 			}
 		} else {
-			// TODO use context
-			groupParticipants, err = cli.getBroadcastListParticipants(to)
-			if err != nil {
-				err = fmt.Errorf("failed to get broadcast list members: %w", err)
-				return
+			if len(req.Participants) != 0 {
+				groupParticipants = req.Participants
```
HafidzEvermos, line 311: IMHO it's better to validate/eliminate the participants that violate user privacy settings like blacklist
see getBroadcastListParticipants for reference

purpshell:
#900 has effects here, what do you think @devlikepro?

zhamghaoran:
This is a great feature, but it encountered a conflict and wasn't merged into master. Can anyone resolve this issue? 🙏🙏🙏

PR #806, opened 2025-05-08
fix(client): introduce worker pool to prevent node handling bottlenecks

eikarna:
Fixes a performance issue where, upon bot startup as a normal user, incoming message queues could overwhelm the single-threaded loop, causing node handling to take over a minute.

This introduces a worker pool to parallelize message processing and prevent long blocking times during message queue draining.

Example warning log:
04:43:48.190 [Client WARN] Node handling took 1m15.00855867s for <message from="phone:serverid@s.whatsapp.net" id="...">

### DIFF
--- a/client.go
+++ b/client.go
@@ -733,34 +733,51 @@ func stopAndDrainTimer(timer *time.Timer) {
 }
 
 func (cli *Client) handlerQueueLoop(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


tulir:
Nodes must be handled in order, but you can always have your own event handler spawn a goroutine instead of blocking if you don't care about message order

PR #824, opened 2025-05-21
client: check downloaded file size only when fileLength >= 0

rnons:
similar to https://github.com/tulir/whatsmeow/blob/106fc9aa085a904cd5a12f284a7bad254399c520/download.go#L291


tulir (approved):
Huh, that original check is 4 years old, I wonder how it got lost in the new download to file stuff 🤔

PR #830, opened 2025-05-29
client: stop reconnecting on expected disconnect

hifi:
If we are requested to disconnect during auto-reconnect wait we will immediately cancel it.

There's still a tiny gap in the reconnect loop that it can hit connect after expected disconnect, could use unlockedConnect and locks around it to check that we weren't disconnected just before.

Posting the PR for feedback now if this is worth all the added complexity.


client.go
```diff
@@ -66,7 +66,8 @@ type Client struct {
 	wsDialer   *websocket.Dialer
 
 	isLoggedIn            atomic.Bool
-	expectedDisconnect    atomic.Bool
+	expectedDisconnectMu  sync.Mutex
+	expectedDisconnectC   chan struct{}
```
tulir, line 70: This smells like an `exsync.Event`

client.go
```diff
@@ -425,7 +429,14 @@ func (cli *Client) SetWSDialer(dialer *websocket.Dialer) {
...
 func (cli *Client) Connect() error {
-	err := cli.connect()
+	if cli == nil {
+		return ErrClientIsNil
+	}
+
+	cli.socketLock.Lock()
+	defer cli.socketLock.Unlock()
+
+	err := cli.unlockedConnect()
```
tulir, line 439: What's this refactor for? Does it make any difference?

hifi: Shoot, that's a leftover from a previous take.

PR #833, opened 2025-06-01
fix: duplicate table whatsmeow_event_buffer

KAnggara75:
fix duplicate tabele name whatsmeow_event_buffer

`panic: failed to upgrade database: failed to run upgrade v8->v9: ERROR: relation "whatsmeow_event_buffer" already exists (SQLSTATE 42P07) [recovered]`

`panic: failed to upgrade database: failed to run upgrade v8->v9: ERROR: relation "whatsmeow_event_buffer" already exists (SQLSTATE 42P07)`

please check 
https://github.com/tulir/whatsmeow/blob/main/store/sqlstore/upgrades/00-latest-schema.sql
and
https://github.com/tulir/whatsmeow/blob/main/store/sqlstore/upgrades/09-decryption-buffer.sql

there is have same sql to create new table whatsmeow_event_buffer,

### DIFF
--- a/store/sqlstore/upgrades/00-latest-schema.sql
+++ b/store/sqlstore/upgrades/00-latest-schema.sql
@@ -141,13 +141,3 @@ CREATE TABLE whatsmeow_lid_map (
 	lid TEXT PRIMARY KEY,
 	pn  TEXT UNIQUE NOT NULL
 );
-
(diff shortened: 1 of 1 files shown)


tulir:
There's no duplicate. The latest schema is for new databases, it creates all tables and jumps straight to v9 (as per the first line). The other files are incremental upgrades for old databases

PR #837, opened 2025-06-04
I've updated the privacy settings for "Selected contact blacklist" an…

coresen:
I've updated the privacy settings for "Selected Contacts" and can now access full privacy details.

### DIFF
--- a/group.go
+++ b/group.go
@@ -8,6 +8,7 @@ package whatsmeow
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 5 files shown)


coresen:
fix receipt

devlikepro:
`DHash` looks like super internal thing that shouldn't be in public API like `SetPrivacySetting` :thinking:

coresen:
> `DHash` looks like super internal thing that shouldn't be in public API like `SetPrivacySetting` 🤔

I get that it's probably meant to act like a timestamp, but the catch is, you *have to* submit the Dhash from the last change when you update, otherwise it just doesn't work.

coresen:
Add community owner update

coresen (commented):
Add community owner update

PR #851, opened 2025-06-18
fix: correct SetGroupDescription to properly handle deletes and updates

xTudoS:
This PR add missing attributes for description updates/deletions


purpshell (approved):
LGTM

Not sure why WhatsApp is doing this.. 🤔 Maybe it is to sequence these updates properly, and also probably just to track these changes (as always)

xTudoS:
> LGTM
> 
> Not sure why WhatsApp is doing this.. 🤔 Maybe it is to sequence these updates properly, and also probably just to track these changes (as always)

I think its a way to deprecate old devices and unofficial apps

proto/waUserPassword/WAProtobufsUserPassword.proto
purpshell, line 1: unrelated proto change

proto/waUserPassword/WAProtobufsUserPassword.pb.go
purpshell, line 1: unrelated proto change

proto/waStatusAttributions/WAWebProtobufsStatusAttributions.proto
purpshell, line 1: unrelated proto change

group.go
```diff
@@ -1052,16 +1052,37 @@ func (cli *Client) SetGroupMemberAddMode(ctx context.Context, jid types.JID, mod
...
+	var contentNodes []waBinary.Node
+    if description != "" {
+        contentNodes = append(contentNodes, waBinary.Node{
+            Tag:     "body",
+            Content: []byte(description),
+        })
+    }
+
+    content := waBinary.Node{
+        Tag:     "description",
```
purpshell, line 1081: I would look further to see if this is change that also affects topics etc. not just descriptions 👀

xTudoS: Are you going to try? This PR is almost 1 year old

PR #855, opened 2025-06-27
fix: message parsing when 'from' field uses LID format

flavio-evocorp:
whatsapp-log-security

I identified an issue in the mapping of some private contacts with numbers in the @lid format.

After enabling debug logs, we discovered that some numbers are arriving with a different pattern in the from, sender_pn, and sender_lid fields.

**Most of the time, as shown in LOG 01:**
```
from=number_of_contact@s.whatsapp.net
sender_lid=numberid@lid
```

**In some cases, however, the pattern is like LOG 02:**
```
from=numberid@lid
sender_pn=number_of_contact@s.whatsapp.net
```

The fix implemented in this pull request should resolve cases where the message comes in the pattern shown in LOG 02.

### DIFF
--- a/message.go
+++ b/message.go
@@ -123,8 +123,15 @@ func (cli *Client) parseMessageSource(node *waBinary.Node, requireParticipant bo
 			source.Chat = from
 		}
 	} else {
-		source.Chat = from.ToNonAD()
-		source.Sender = from
+		if from.Server == types.HiddenUserServer {
+			senderPn := ag.OptionalJIDOrEmpty("sender_pn")
+			source.Chat = senderPn.ToNonAD()
+			source.Sender = senderPn
+		} else {
+			source.Chat = from.ToNonAD()
+			source.Sender = from
+		}
+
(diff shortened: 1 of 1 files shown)


tulir:
The current parsing is correct. The phone number can be found in the SenderAlt field

flavio-evocorp:
When the log comes in the format of LOG 02 (as shown in the pull request image), the `SenderAlt` field is left empty because, where the mapping is done, the `AddressingMode` field is always coming as empty.

In this code snippet, it always falls into the `else` block:

```go
source.Chat = from.ToNonAD()
source.Sender = from

if source.AddressingMode == types.AddressingModeLID {
    source.SenderAlt = ag.OptionalJIDOrEmpty("sender_pn")
} else {
    // THIS BLOCK IS ALWAYS BEING EXECUTED, EVEN WHEN FROM IS A @lid
    source.SenderAlt = ag.OptionalJIDOrEmpty("sender_lid")
}
```

I've been testing the change I made to the code in production for two days, and it's working for both cases.

Issue #866, opened 2025-07-08
GetBusinessProfile panics with type assertion error: []binary.Node vs []uint8


aldinokemal:
## Bug Description

The `GetBusinessProfile` function in whatsmeow panics with a type assertion error when trying to parse business profile XML responses. The library receives valid XML data but fails to parse it correctly.

## Error Details

**Panic Message:**
```
panic: failed to parse business profile: the profile data contains unsupported format or this account may not be a business account
```

**Underlying Error:**
```
interface conversion: interface {} is []binary.Node, not []uint8
```

## Environment

- **Go Version:** 1.24.0
- **whatsmeow Version:** `v0.0.0-20250707142839-10067399df1f`
- **Platform:** macOS (Darwin 24.5.0)

## XML Response Data

The XML response that causes the parsing failure:

```xml
<iq from="s.whatsapp.net" id="249.123-6" type="result">
<business_profile>
<profile jid="628112640770@s.whatsapp.net" tag="125605989">
<address>Jl. Pandanaran No.252d, Recosari, Banaran, Kec. Boyolali, Kabupaten Boyolali, Jawa Tengah 57313, Indonesia.                                                           https://maps.app.goo.gl/mz4BLgpREVfvEw8o9</address>
<description><\!-- 493 bytes --></description>
<email>doozboyolali@gmail.com</email>
<website>https://www.instagram.com/doozoptik?utm_source=qr&igsh=ZDR1M3IxbTVqYnpk</website>
<latitude>-7.5478</latitude>
<longitude>110.5805</longitude>
<business_hours timezone="Asia/Jakarta">
<business_hours_config close_time="1200" day_of_week="sun" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="mon" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="tue" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="wed" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="thu" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="fri" mode="specific_hours" open_time="600"/>
<business_hours_config close_time="1200" day_of_week="sat" mode="specific_hours" open_time="600"/>
</business_hours>
<categories>
<category id="145118935550090">Medical & health</category>
</categories>
<profile_options>
[...]

PR #559, opened 2024-03-27
feature: implement query business profile

stephentth:
adding new api for query info of a business profile

debug log
```
16:35:03.410 [Client/Send DEBUG] <iq id="27.32-3" to="s.whatsapp.net" type="get" xmlns="w:biz"><business_profile v="244"><profile jid="1234567890@s.whatsapp.net"/></business_profile></iq>
16:35:03.908 [Client/Recv DEBUG] <iq from="s.whatsapp.net" id="27.32-3" type="result">
  <business_profile>
    <profile jid="1234567890@s.whatsapp.net" tag="2815578256">
      <address>1 Stadium Dr, Singapore 397629</address>
      <email>qwerty@gmail.comm</email>
      <business_hours timezone="Asia/Singapore">
        <business_hours_config close_time="0" day_of_week="mon" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="tue" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="wed" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="thu" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="fri" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="sat" mode="open_24h" open_time="0"/>
        <business_hours_config close_time="0" day_of_week="sun" mode="open_24h" open_time="0"/>
      </business_hours>
      <categories><category id="1022050661163852">Finance</category></categories>
      <profile_options>
        <commerce_experience>none</commerce_experience>
        <cart_enabled>true</cart_enabled>
        <direct_connection>false</direct_connection>
      </profile_options>
      <direct_connection enabled="false"><features name="default"/></direct_connection>
      <automated_type>unknown</automated_type>
    </profile>
  </business_profile>
</iq>


(dev log)
[...]

PR #867, opened 2025-07-08
feat: Add comprehensive XML parsing tests for business profile

aldinokemal:
## Summary

This PR adds comprehensive XML parsing tests for business profile functionality to address the panic described in issue #866 and https://github.com/tulir/whatsmeow/pull/559#issuecomment-2958011609

## Changes

- *Added *: Tests parsing business profile from a structure that exactly matches the XML response from issue #866

## Issue Resolution

Addresses #866 where the  function would panic with:
`panic: failed to parse business profile: the profile data contains unsupported format or this account may not be a business account`

## Screenshoot for evidence
Have tested in my local from fork repo

image


aldinokemal:
Hi @tulir , 
business day is missing, I believe it should `day_of_week` instead of `dow`
image

aldinokemal:
I suggest adding a fallback in case you're unsure about my payload in the issue.

```go
dow := config.AttrGetter().String("dow")
if dow == "" {
    dow = config.AttrGetter().String("day_of_week")
}
```

PR #894, opened 2025-08-05
feat: add LID (Local ID) support to IsOnWhatsApp and GetUserInfo functions

DavidsonGomes:
- Add LID field to IsOnWhatsAppResponse struct
- Add LID field to UserInfo struct
- Implement LID Store lookup in IsOnWhatsApp function
- Implement LID Store lookup in GetUserInfo function
- LID format: xxxxx@lid (different from phone number)
- LID is retrieved from local database when available
- Backwards compatible - LID field is optional (*JID)

### DIFF
--- a/types/user.go
+++ b/types/user.go
@@ -24,6 +24,7 @@ type UserInfo struct {
 	Status       string
 	PictureID    string
 	Devices      []JID
+	LID          *JID // The local ID (if available)
 }
 
 type BotListInfo struct {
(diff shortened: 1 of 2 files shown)


tulir:
These changes look unnecessary, you can call the store yourself if you need the LID

What should be done is either a new method or adjusting GetUserInfo to ask the server to return LIDs when a phone number is provided, but that requires checking how official whatsapp web does it.

PR #896, opened 2025-08-08
feat(client): Introduce platform abstractions

jlucaso1:
This pull request prepares the `whatsmeow` library for use in WebAssembly environments by refactoring key platform-specific dependencies into abstract interfaces. The goal of this change is to enable the creation of a separate `whatsmeow-wasm` adapter/package without requiring any further modifications to the core library.
Another benefit is the possibility to change the websocket client (gorilla), because now is decoupled (this is useful for testing)

**Summary of Changes**

The primary motivation is to decouple logic that cannot run in a browser's WASM sandbox or other platforms with:

1.  **WebSocket Dialing:** The client's dependency on `gorilla/websocket` has been moved behind a new `iface.WebSocketDialer` interface.
    *   The `Client` now uses this interface for all dialing operations.
    *   For 100% backward compatibility, `NewClient` initializes with a `net.GorillaDialer` that uses the original library.
    *   A new public method, `client.SetWSDialer()`, is provided as a clean injection point for a WASM-compatible dialer.

2.  **File System Operations:**
    *   **`fallocate`:** File pre-allocation is now handled by a platform-agnostic `fallocate.Fallocate` function. The native implementation (`!wasm` build tag) uses the original `go.mau.fi/util/fallocate`, preserving performance benefits.
    *   **Temporary Files:** `UploadReader` no longer calls `os.CreateTemp`. It now requires the caller to provide an `io.ReadWriteSeeker` for use as a temporary buffer, making the function fully platform-agnostic. ⚠️ This can be a potential breaking change, because we dont create the folder anymore.

**Backward Compatibility and Non-Breaking Design**

This PR was designed to be as non-intrusive as possible:

*   ✅ **No Breaking API Changes:** All public methods retain their signatures and behavior for existing users. (maybee except the temporary files that i mentioned earlier)
*   ✅ **Proxy Functionality Preserved:** `SetProxy` and `SetSOCKSProxy` continue to work seamlessly for native builds by configuring the default `GorillaDialer`. This logic is excluded from WASM builds using build tags.

**How to Use for WASM** (you can check an experiment in: https://github.com/jlucaso1/whatsmeow/tree/wasm)

With these changes, a separate `whatsmeow-wasm` package can now be created. That package would:
1.  Implement the `iface.WebSocketDialer` interface using JS interop to call the browser's `WebSocket` API.
2.  Provide an in-memory `io.ReadWriteSeeker` (like `&bytes.Buffer{}`) to the `UploadReader` function.
3.  Initialize the client and use `client.SetWSDialer()` to inject its browser-based dialer.

This architectural improvement ensures a clean separation of concerns and makes `whatsmeow` extensible to new platforms without cluttering the core library.

### DIFF
--- a/client.go
+++ b/client.go
@@ -19,7 +19,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/gorilla/websocket"
 	"go.mau.fi/util/exhttp"
 	"go.mau.fi/util/exsync"
 	"go.mau.fi/util/random"
@@ -27,6 +26,8 @@ import (
 
 	"go.mau.fi/whatsmeow/appstate"
 	waBinary "go.mau.fi/whatsmeow/binary"
+	"go.mau.fi/whatsmeow/iface"
+	wanet "go.mau.fi/whatsmeow/net"
 	"go.mau.fi/whatsmeow/proto/waE2E"
 	"go.mau.fi/whatsmeow/proto/waWa6"
 	"go.mau.fi/whatsmeow/proto/waWeb"
@@ -65,7 +66,7 @@ type Client struct {
 	socket     *socket.NoiseSocket
 	socketLock sync.RWMutex
 	socketWait chan struct{}
-	wsDialer   *websocket.Dialer
+	wsDialer   iface.WebSocketDialer
 
 	isLoggedIn            atomic.Bool
 	expectedDisconnect    *exsync.Event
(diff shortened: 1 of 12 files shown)


tulir:
Everything here looks unnecessary:

* No point in abstracting websockets, it should just be switched to https://github.com/coder/websocket completely without any abstractions
* The fallocate change is pointless, go-util already has build tags and is a no-op on unsupported platforms https://github.com/mautrix/go-util/blob/main/fallocate/fallocate_unknown.go
* The `UploadReader` change looks like it should just be a comment mentioning passing a temp file is required on platforms which don't support creating files via `os`, there's no benefit to removing the fallback

jlucaso1:
@tulir thank you for the feedback. I didn't know about this websocket lib. Is there a chance it could become standard for this project? I saw that it has some interesting advantages.
And can you provide a code to switch the libraries (gorilla to this new one)? (I don't know how)

tulir:
That is the plan yes. The gorilla suite has been mostly unmaintained since 2022. There was an attempt to revive it, but the new maintainers made a mess of that with several bad changes in the websocket library (they had to make a new release reverting all their changes: https://github.com/gorilla/websocket/releases/tag/v1.5.3 - there haven't been any releases since then)

jlucaso1:
@tulir I've removed the unecessary abstractions (except the websocket interface)
I've changed the gorilla in favor of the coder implementation. Can you review again?

jlucaso1:
@tulir nice work

PR #898, opened 2025-08-08
[BRI-31699] Argo Support for Newsletter Mex IQs

purpshell:
This adds support for the Argo protocol for newsletter mex iQs when running as desktop. Note that this Argo format might be more expansive and probably includes business features, there needs to be more testing. Also note that the list of ids is not complete when it comes to the newsletter operations, and that some work needs to be done there as well.


newsletter.go
```diff
@@ -10,10 +10,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
```
tulir, line 13: This is not allowed

newsletter.go
```diff
@@ -118,9 +124,53 @@ const (
 	mutationCreateNewsletter       = "6234210096708695"
 	mutationUnfollowNewsletter     = "6392786840836363"
 	mutationFollowNewsletter       = "9926858900719341"
+
+	// desktop & mobile
+	queryFetchNewsletterDesktop        = "9779843322044422"
```
tulir, line 129: Are the query IDs really different 🤔
The web here are probably outdated

purpshell: It looks like they are btw, but I found no reference to the desktop ones in the web version

argo/argo.go
```diff
@@ -14,22 +13,21 @@ var Store map[string]wire.Type
 
 var QueryIDToMessageName map[string]string
 
+//go:embed argo-wire-type-store.argo
+var wireTypeStoreBytes []byte
+
+//go:embed name-to-queryids.json
+var jsonMapBytes []byte
+
 func init() {
```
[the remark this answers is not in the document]
purpshell: Then the `Init()` will be called inside the newsletter file in the logic that needs the store?

argo/argo.go
```diff
@@ -0,0 +1,68 @@
...
+		QueryIDToMessageName = m
+	})
+	return initErr
+}
+
+func MustInit() {
+	if err := Init(); err != nil {
+		panic(err)
+	}
+}
```
tulir, line 55: What's this for? 🤔 If it's not used it shouldn't be added

```suggestion
```

newsletter.go
```diff
@@ -151,14 +201,40 @@ func (cli *Client) sendMexIQ(ctx context.Context, queryID string, variables any)
 	if !ok {
 		return nil, fmt.Errorf("unexpected content type %T in mex response", result.Content)
 	}
-	var gqlResp types.GraphQLResponse
-	err = json.Unmarshal(resultContent, &gqlResp)
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal graphql response: %w", err)
-	} else if len(gqlResp.Errors) > 0 {
-		return gqlResp.Data, fmt.Errorf("graphql error: %w", gqlResp.Errors)
+	if result.AttrGetter().String("format") == "argo" {
```
tulir, line 204: Is `format` present always? If not, `OptionalString` is more appropriate

```suggestion
	if result.AttrGetter().OptionalString("format") == "argo" {
```

tulir (approved):
Otherwise looks good

PR #925, opened 2025-09-10
message,send: share/request pn from business

purpshell:
This PR allows whatsmeow users (mainly business running Meta Ads) to receive the phone numbers of users that are now appearing as LID. It also allows you to share your phone number with a business if needed.

This might appear later on in more normal WhatsApp uses, as WhatsApp is gearing towards username based and LID based identification, leaving phone numbers to eventually become a privacy bound identifier.

### DIFF
--- a/message.go
+++ b/message.go
@@ -751,6 +751,11 @@ func (cli *Client) handleProtocolMessage(ctx context.Context, info *types.Messag
 		cli.storeLIDSyncMessage(ctx, protoMsg.GetLidMigrationMappingSyncMessage().GetEncodedMappingPayload())
 	}
 
+	if protoMsg.GetType() == waE2E.ProtocolMessage_SHARE_PHONE_NUMBER {
+		// user shared phone number with a business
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -751,6 +751,11 @@ func (cli *Client) handleProtocolMessage(ctx context.Context, info *types.Messag
 		cli.storeLIDSyncMessage(ctx, protoMsg.GetLidMigrationMappingSyncMessage().GetEncodedMappingPayload())
 	}
 
+	if protoMsg.GetType() == waE2E.ProtocolMessage_SHARE_PHONE_NUMBER {
+		// user shared phone number with a business
+		cli.StoreLIDPNMapping(ctx, info.Sender, info.SenderAlt)
```
tulir, line 756: This also seems redundant since it already happens in `handleEncryptedMessage`

tulir:
Doesn't look like either of the changes are needed

PR #947, opened 2025-09-28
user: Add support for LID in USync Protocol + Store mappings

purpshell:
The aim of this PR is to add support for LIDs in the USyncProtocol calls.

The USync protocol allows you to get the LID of any PN JID, not sure why this hasn't been implemented before. 

What I suggest happens next is a way to allow LIDForPN calls to somehow fallback to this as a backup, rather than being empty-handed.


purpshell:
> What I suggest happens next is a way to allow LIDForPN calls to somehow fallback to this as a backup, rather than being empty-handed.

This is what we are doing in Baileys.

You can definetely just setup a minimalist USyncQuery with just the LID protocol tag, and put up to 500 users in 1 request (I recommend maximum 100 per request, whatsapp has some dark wizardry on Web that allows 500).

purpshell:
> > What I suggest happens next is a way to allow LIDForPN calls to somehow fallback to this as a backup, rather than being empty-handed.
> 
> This is what we are doing in Baileys.
> 
> You can definetely just setup a minimalist USyncQuery with just the LID protocol tag, and put up to 500 users in 1 request (I recommend maximum 100 per request, whatsapp has some dark wizardry on Web that allows 500).

Upon further confirmation, we can do up to 500 a time with 20 seconds in between

CleanShot 2025-09-28 at 07 14 51

tulir (approved):

PR #952, opened 2025-10-02
send: Acquire LIDs from PNs by USync during message enc

purpshell:
This PR is based off the already existing #947, all it does is experiment with fetching missing LIDs from the usync protocol


purpshell:
UPDATE: The only real considerations for this seem to be batching the userInfo requests in amounts of 500 for every 20 seconds each. Not sure if this is something that we need to implement or not, maybe this could be an issue in large groups (though not sure if that is even a consideration for this logic here)

PR #953, opened 2025-10-02
message,receipt: Fix message <ack>s and support peer_msg receipts

purpshell:
WhatsApp Web is no longer sending <ack>s for successfully received messages. The existence of this node is a valid fingerprinting vector.

I ask people to test the PR to see if there would be any kind of accumulation of messages or issues in socket stability after the changes in the PR.


purpshell:
Just tested these changes: 

Receiving messages is successful, and upon reconnect, there are no retried messages (with an offline counter).

Also didn't get immediately banned 😬

PR #967, opened 2025-10-11
business: add support for fetching order details

salman0ansari:
- added function `GetOrderDetails`  to fetch order details 
- a new `business.go` file is introduced to house this and other business-related logic.
- data structures for the order details are defined in `types/business.go`.

### Usage

```go
// you can get order id and token from the OrderMessage
orderDetails, err := client.GetOrderDetails(context.Background(), "order-id-here", "order-token-here")
if err != nil {
    // handle error
}
fmt.Printf("Order %s has %d products\n", orderDetails.ID, len(orderDetails.Products))
```

### Response 

```json
{
  "id": "order-id-here",
  "created_at": "YYYY-MM-DDTHH:MM:SS+TZ",
  "catalog_id": "catalog-id-here",
  "price": {
    "subtotal": 1,
    "total": 1,
    "currency": "USD",
    "price_status": "provided"
  },
  "products": [
    {
      "id": "product-id-here",
      "image_id": "image-id-here",
      "image_url": "https://example.com/product-image.jpg",
      "price": 1,
      "currency": "USD",
      "name": "Product Name",
      "quantity": 1,
      "variant_info": {}
    }
  ]
}
```


hrizal:
nice ...

Diden05:
@tulir merge?

purpshell (approved):
verified working on whatsapp business android

business.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Helper to get the string content of a child node.
+func getStringChild(node waBinary.Node, tag string) string {
+	child, ok := node.GetOptionalChildByTag(tag)
+	if !ok {
+		return ""
+	}
+	content, _ := child.Content.([]byte)
+	return string(content)
+}
```
purpshell, line 64: Not the right place for a helper like this.

salman0ansari: what should be done?

salman0ansari: should i keep this scoped to the order-details parser instead ??

```go
func parseOrderDetailsNode(orderNode waBinary.Node) (*types.OrderDetails, error) {
	getStringChild := func(node waBinary.Node, tag string) string {
		child, ok := node.GetOptionalChildByTag(tag)
		if !ok {
			return ""
		}
		content, _ := child.Content.([]byte)
		return string(content)
	}

	ag := orderNode.AttrGetter()
}

```

PR #970, opened 2025-10-15
prekeys: update total prekey count to 812

purpshell:


purpshell:
Logic should be to reach 812, not to batch on each request 812 keys. Even though I see sometimes on WA Web pre-keys in the 1900s etc.

PR #971, opened 2025-10-15
pair: handle connection pair props

purpshell:


tulir:
I thought one of the history sync payloads had the timestamp

pair.go
```diff
@@ -93,9 +96,13 @@ func (cli *Client) handlePairSuccess(node *waBinary.Node) {
 	jid, _ := pairSuccess.GetChildByTag("device").Attrs["jid"].(types.JID)
 	lid, _ := pairSuccess.GetChildByTag("device").Attrs["lid"].(types.JID)
 	platform, _ := pairSuccess.GetChildByTag("platform").Attrs["name"].(string)
+	clientPropsBytes := pairSuccess.GetChildByTag("client-props").Content.([]byte)
```
tulir: This shouldn't panic if it's missing

```suggestion
	clientPropsBytes, _ := pairSuccess.GetChildByTag("client-props").Content.([]byte)
```

pair.go
```diff
@@ -93,9 +96,13 @@ func (cli *Client) handlePairSuccess(node *waBinary.Node) {
...
+
+	var props waCompanionReg.ClientPairingProps
+	err := proto.Unmarshal(clientPropsBytes, &props)
```
tulir: This error seems to be ignored. `_` should be used instead if the error doesn't matter, but alternatively it could be logged

tulir (commented):
The CI failure also needs to be fixed by listening to the linter's opinions

tulir:
> I thought one of the history sync payloads had the timestamp

Does the official web app starts sending LID DMs from either flag being present? or do both flags always get set at the same time?

purpshell:
> > I thought one of the history sync payloads had the timestamp
> 
> Does the official web app starts sending LID DMs from either flag being present? or do both flags always get set at the same time?

I'll test this now

PR #978, opened 2025-10-23
download: Fail faster on invalid download

radon-at-beeper:
If the user passes in an empty or malformed string for the path, rather than doing string concatenation with it and then trying repeatedly to resolve a nonexistent domain, just error out immediately, as it's obviously not a valid parameter.


radon-at-beeper:
Derp yes.

PR #998, opened 2025-11-13
group: parse membership request notifications

guilhermejansen:
- parse `created|deleted|revoked_membership_requests` inside `group.go` using the same member parser used for add/remove/promote/demote
- expose the parsed JIDs via new `events.GroupInfo` fields together with the request method, keeping all other handlers untouched
- verified against real traffic where membership requests previously entered `UnknownChanges`

- replayed group notifications that contained `created_membership_requests` and `revoked_membership_requests` and confirmed the new fields are populated while other updates (join/leave/promote) still behave as before


purpshell (commented):
rebase your commits properly.

PR #999, opened 2025-11-13
call: add dispatch for the entire call packets in a generic event

Darkratos:
This adds an unified way to get all call packets within a single handler


purpshell (commented):
interesting change but we should probably decrypt the offer ourselves on the library side.

Darkratos:
> interesting change but we should probably decrypt the offer ourselves on the library side.

Yes, but this is just a way to unify all events into one place for easier handling, similar to CB:call from Baileys.

Once I'm more acostumed with the library I can help put up the decryption process

tulir:
Probably won't add this, although might be reasonable to have a method in `DangerousInternals()` that allows subscribing to arbitrary nodes

PR #1003, opened 2025-11-16
Change link_code_pairing_nonce content type

tovmeod:
this was not working for me so I compared with baileys, this is how they implemented, I built and tested locally and worked


tulir:
`[]byte{0}` is the correct value

image

tovmeod:
I'm not sure where is this, I wasted a week trying to pair and with this change it worked for me finally.
for comparison Baileys use '0'
https://github.com/Itsukichann/Baileys/blob/0bff504b55f11d50f6f00bc6857914e326acd98e/lib/Socket/socket.js#L648C1-L650C41

tulir:
I'd guess you were using an outdated version of whatsmeow and only updated it when you tested with your new branch

PR #1021, opened 2025-12-01
feat(group): Added support for group pause status

zhamghaoran:
- Added a new Suspended field to the group event structure in events.go
- Handling the suspended state when parsing binary nodes in group.go
- Set EVT when the group is paused. Suspended is true
- Update the unknown change handling logic to accommodate the new fields


tulir:
Sounds like something that should be parsed when getting group info too, not only in events 🤔

zhamghaoran:
> Sounds like something that should be parsed when getting group info too, not only in events 🤔

Getting group information will also go to the parseGroupNode function

PR #1058, opened 2026-01-12
fix: add recover to handleRetryReceipt goroutine to prevent panics from crashing client

AlejoGarat:
Problem
* A panic in handleRetryReceipt (e.g., nil pointer in libsignal) can crash the entire WhatsMeow client.
* This is especially problematic for retry receipts where the library may occasionally panic.

Solution:
* Wrap the goroutine that calls handleRetryReceipt in a defer + recover.
* Log the panic instead of letting it crash the process.


zhamghaoran:
I've encountered the same problem.

PR #1059, opened 2026-01-13
Optimize the message sending chain to prevent database deadlock caused by session updates.

zhamghaoran:
- Changed synchronous session storage to asynchronous execution and added a panic recovery mechanism
- Modified the session cache storage method from batch storage to individual storage to avoid deadlocks
- Added detailed session storage failure logging to support continuing to process other sessions when a single session fails
- Unified asynchronous processing of session cache storage logic in send.go and sendfb.go
- Remove concurrency issues that may be caused by batch operations and improve system stability

### DIFF
--- a/send.go
+++ b/send.go
@@ -1306,10 +1306,12 @@ func (cli *Client) encryptMessageForDevices(
 		}
 	}
 	go func() {
-		err := recover()
-		if err != nil {
-			cli.Log.Warnf("panic occurred in PutCachedSessions: %v", err)
-		}
+		defer func() {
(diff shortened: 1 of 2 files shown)


tulir:
This is incorrect, sessions must be saved synchronously

zhamghaoran:
I believe that not saving synchronously might lead to session renegotiation, but the problem I'm encountering is that synchronous saving causes a database deadlock when I send messages to many groups simultaneously. Therefore, I want to save asynchronously.

PR #1062, opened 2026-01-23
retry: fix waiting for message

rbtyuki:
When the receipt retry event includes a LID, it should be converted to a PN to retrieve the message from the cache.


tulir:
Looks like a real issue, thanks for spotting it. It's definitely not going to fix all cases of "waiting for message" and probably won't even fix a majority of them, because this case doesn't affect groups at all. The fix here is incomplete too: the other direction needs to be handled as well (receiving a retry receipt to your phone number when your account is already sending LID DMs). Also, ignoring errors isn't allowed in most cases

PR #1069, opened 2026-01-29
fix(crypto): retry decryption with stored LIDs/PNs for poll votes

jvkabum:
### Summary
This PR implements a robust retry mechanism for decrypting poll votes and secret messages to fix persistent `cipher: message authentication failed` errors. These errors occur due to mismatches between the sender JID derived from the message key (e.g., PN) and the sender JID stored in the database (e.g., LID).

### Changes
- **[msgsecret.go](cci:7://file:///d:/whatsmeow/msgsecret.go:0:0-0:0)**: Modified [decryptMsgSecret](cci:1://file:///d:/whatsmeow/msgsecret.go:92:0-117:1) to attempt decryption using both the sender JID from the message key and the stored sender JID. If the first attempt fails, it retries with the alternative JID before returning an error.
- **[store/sqlstore/store.go](cci:7://file:///d:/whatsmeow/store/sqlstore/store.go:0:0-0:0)**: Updated `getMsgSecret` query to correctly map LIDs to PNs (and vice-versa) using `whatsmeow_lid_map` when searching for message secrets, ensuring retrieval regardless of the stored JID format.

### Motivation
Fixes decryption failures for poll votes where the poll creator's JID format differs from the session storage, a common issue with recent LID updates.

### DIFF
--- a/msgsecret.go
+++ b/msgsecret.go
@@ -94,23 +94,27 @@ func (cli *Client) decryptMsgSecret(ctx context.Context, msg *events.Message, us
 	if cli == nil {
 		return nil, ErrClientIsNil
 	}
-	origSender, err := getOrigSenderFromKey(msg, origMsgKey)
+	origSenderFromKey, err := getOrigSenderFromKey(msg, origMsgKey)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


tulir:
This seems to be a duplicate of #991

jvkabum:
alteraçao nao vinculada ao projeto main

PR #1072, opened 2026-02-08
fix : SenderAlt empty if IsFromMe is True in MessageSource

ZhanNexus:


tulir:
the data in MessageSource is just what the server sends, it won't be filled with local data

Issue #1077, opened 2026-02-14
Send message to large groups take long time than baileys


jeffersonfelixdev:
When I send messages to large groups (about 700 to 1000 participants) the message take about 30 seconds to send.

Sending a message to same group take no more than 2 secs using Baileys.

MarceloGouveia77:
i don't have this problem here with large groups.

working with 80 groups with mean of 800 participants per group and everything works well and very fast

jeffersonfelixdev:
I am using WUZAPI (https://github.com/asternic/wuzapi), maybe there is a problem with their implementation.

Does whatsmeow have group caching like baileys?

MarceloGouveia77:
you can make your own cache, i've made:

func (mycli *MyClient) getGroupInfoCached(ctx context.Context, jid types.JID) (*types.GroupInfo, error)

so instead of fetch from server everytime, i get from cache. i did the same with some stuffs, like images (for example sending same image to all groups):

```
type CachedImage struct {
	Upload    whatsmeow.UploadResponse
	MimeType  string
	Thumbnail []byte
}
```

I think that helped my implementation run very quickly.

jeffersonfelixdev:
Investigating wuzapi code, I didn't find any function that fetch group info before send message.

Btw, I will try to create my own cache

jeffersonfelixdev:
I figured out that the long time occurs at this point:

```go
resp, err = clientManager.GetWhatsmeowClient(txtid).SendMessage(context.Background(), recipient, msg, whatsmeow.SendRequestExtra{ID: msgid})
```

Ali-hmede-codes:
does whatsmeow is good to send a bulk message to WhatsApp groups in a short time ?

ahmedRSA:
any solution ?

ralfiannor:
Do you use PostgreSQL or SQLite?

ahmedRSA:
> Do you use PostgreSQL or SQLite?

this is not db problem 

its whatsmeow problem, same problem happen with when you send status and if you have huge contact list it sometimes takes 5-6mins

idk why tulir rufuses to work on this even after years 
 some issues/pr relaetd to perfermance has been open since 2025 #900 #800 #721

deemwario:
The group-metadata caching people are trying here (getGroupInfoCached, image upload caching) will help, but it's not the dominant cost at 700–1000 participants — worth cross-linking this to #900, which actually profiled the send path rather than guessed at it.

[...]

PR #1080, opened 2026-02-15
fix: parallel encryption and caching to send messages to large groups

jeffersonfelixdev:
# Performance: faster message sending to large groups

## Problem

Sending messages to groups with 700+ participants could take over 30 seconds due to:

1. **Single large usync** – One `GetUserDevices` call with hundreds of JIDs in a single request, leading to large payloads and higher timeout risk.
2. **Sequential encryption** – The message was encrypted for each participant device one after another; with 700+ devices this became a long sequential chain of crypto operations.
3. **Unbounded prekey fetch** – All devices without an existing session were requested in one IQ, which could be slow or hit server limits.

### 1. Batch usync in `GetUserDevices` (`user.go`)

- Split the JID list into chunks of **256** and perform one usync per chunk.
- Keeps each request smaller and more reliable for large groups.

### 2. Parallel device encryption (`send.go`, `sendfb.go`)

- Run encryption for each device in **parallel** using a worker pool (`NumCPU()*2`, cap 32).
- Participant node order is preserved for protocol compatibility.
- Significantly reduces total encryption time on multi-core machines (e.g. ~10x with 8 cores in typical cases).

### 3. Batch prekey fetches (`prekeys.go`)

- Request prekeys in batches of **50** devices instead of a single large IQ.
- Avoids oversized requests and potential server limits or timeouts.

## Testing

- Recommended to test in a **community** or large group (e.g. 700+ participants).
- Use `SendResponse.DebugTimings` to compare before/after: `GetParticipants`, `GetDevices`, `PeerEncrypt`, `Send`.
- Most improvement should show in **PeerEncrypt** due to parallel encryption; **GetDevices** benefits from batching and from cache after the first send.

## Backward compatibility

- No API changes. Behavior and message format are unchanged; only internal performance is improved.
- Cache usage (group cache, device cache) is unchanged and continues to make subsequent sends faster.

### DIFF
--- a/prekeys.go
+++ b/prekeys.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"slices"
 	"time"
 
 	"go.mau.fi/libsignal/ecc"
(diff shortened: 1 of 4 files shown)


jeffersonfelixdev:
Resolves #1077

PR #1082, opened 2026-02-18
Feat wasm compilation

vlmoon99:
I found this library very useful! However, I noticed it didn’t support WASM compilation, so I’ve added that option and provided an example. I’d love to contribute this to the project if you're interested. If you prefer not to have an 'examples' folder, I can easily remove it—just let me know what needs to be changed and I’ll fix it. I’ve tried to cover everything in the example/TUTORIAL_DEPLOY.md file. Please check it out!


tulir:
Definitely not going to add big examples like that. The build tags also don't seem to match coder/websocket

vlmoon99:
> Definitely not going to add big examples like that. The build tags also don't seem to match coder/websocket

I will remove that example; I only included it to demonstrate how the implementation works. Regarding your comment that 'the build tags don't seem to match coder/websocket'—could you clarify what you mean by that?

I created stubs for HTTPClient and HTTPHeader specifically for WASM compilation because those components are unavailable on the JS side. In my view, the library benefits significantly from being able to compile to WebAssembly. Could you briefly explain how you view this approach?"

tulir:
coder/websocket only uses the `js` build tag, not `wasm` https://github.com/coder/websocket/blob/master/dial.go#L1

PR #1084, opened 2026-02-20
fix: handle picture notification when server sends JID/ID (uppercase)

martinsjr:
## Problem

`handlePictureNotification` ignores picture change notifications when the server sends attributes in uppercase (`JID` and `ID` instead of `jid` and `id`). The `AttrGetter` lookup is case-sensitive, so these notifications are discarded with:

Ignoring picture change notification with unexpected attributes: ...

## Solution

- Added helper functions `getJIDFromAttrs` and `getStringFromAttrs` that try multiple attribute keys
- `getJIDFromAttrs` supports both `types.JID` and `string` values (parses with `ParseJID` when needed)
- Explicit validation for required JID and PictureID when applicable

### DIFF
--- a/notification.go
+++ b/notification.go
@@ -70,26 +70,67 @@ func (cli *Client) handleAppStateNotification(ctx context.Context, node *waBinar
 	}
 }
 
+// getJIDFromAttrs tries multiple keys and supports both types.JID and string values.
+func getJIDFromAttrs(attrs waBinary.Attrs, keys ...string) types.JID {
+	for _, k := range keys {
+		if v, ok := attrs[k]; ok && v != nil {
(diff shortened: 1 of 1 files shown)


tulir:
I don't see any evidence of such attributes

zhamghaoran:
This problem does exist; I checked my online logs. @tulir 
[Client DEBUG] Ignoring picture change notification with unexpected attributes: [didn't find required JID attribute 'jid' didn't find required attribute 'id']

tulir:
That's just saying the attribute isn't there, which is known, but there's no way to parse a notification that doesn't have the wanted data. This PR wouldn't fix anything

PR #1098, opened 2026-03-22
fix: dynamically resolve sender JID (PN vs LID) for secret-encrypted messages

Natanel-Shitrit:
### Description
This PR addresses silent decryption failures for `MessageSecret`-encrypted payloads (Poll Votes, Reactions, and Comments) across different chat addressing modes.

**The Issue:**
Previously, encryption functions were hardcoding the sender's identity:
* `EncryptPollVote` hardcoded `cli.getOwnID()` (PN). This caused votes to silently fail to decrypt for other participants in modern groups (`@g.us`) and privacy-enabled chats (`@lid`) because other clients expected the HKDF to be derived from the voter's LID, not their PN.
* `EncryptComment` and `EncryptReaction` hardcoded `cli.getOwnLID()`. While functioning in privacy chats, this breaks them in legacy 1-on-1 PN chats (`@s.whatsapp.net`).

**The Fix:**
* Refactored `encryptMsgSecret` to accept `origMsgInfo *types.MessageInfo` instead of the manually mapped IDs.
* `encryptMsgSecret` now dynamically selects `ownID` (either `cli.getOwnID()` or `cli.getOwnLID()`) by inspecting `origMsgInfo.AddressingMode` and `origMsgInfo.Chat.Server`.
* Updated `EncryptPollVote`, `EncryptComment`, and `EncryptReaction` to route through the centralized logic, resolving the `// TODO is hardcoding LID here correct?` comment.

### Testing Performed
Verified successful payload decryption by other WhatsApp clients (Web and Mobile) in the following environments:
- [x] Standard 1-on-1 chats (`@s.whatsapp.net`)
- [x] Standard Groups (`@g.us`)
- [x] Privacy-enabled/Announcement groups (`@lid`)

**Note:** At this stage, only **Poll Votes** have been thoroughly tested and confirmed working across all three addressing modes. Testing for Reactions and Comments using the newly refactored centralized logic is pending and will be completed in a follow-up.

---

*Disclaimer: Google's Gemini AI was used as an assistive tool to help analyze the root cause, suggest the refactoring, and format this PR. All changes were manually reviewed, understood, applied, and thoroughly double-checked and tested. No AI output was used blindly.*

### DIFF
--- a/internals.go
+++ b/internals.go
@@ -6,747 +6,742 @@
 package whatsmeow
 
 import (
-	"context"
-	"encoding/json"
-	"io"
-	"net/http"
-	"time"
-
-	"go.mau.fi/libsignal/keys/prekey"
-
-	"go.mau.fi/whatsmeow/appstate"
-	waBinary "go.mau.fi/whatsmeow/binary"
-	"go.mau.fi/whatsmeow/proto/waCommon"
-	"go.mau.fi/whatsmeow/proto/waE2E"
-	"go.mau.fi/whatsmeow/proto/waHistorySync"
-	"go.mau.fi/whatsmeow/proto/waMsgApplication"
-	"go.mau.fi/whatsmeow/proto/waMsgTransport"
-	"go.mau.fi/whatsmeow/proto/waServerSync"
-	"go.mau.fi/whatsmeow/socket"
-	"go.mau.fi/whatsmeow/store"
-	"go.mau.fi/whatsmeow/types"
-	"go.mau.fi/whatsmeow/types/events"
-	"go.mau.fi/whatsmeow/util/keys"
+        "context"
+        "encoding/hex"
+        "errors"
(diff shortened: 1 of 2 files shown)


Natanel-Shitrit:
If anyone has a few minutes to help test this, it would be hugely appreciated:

1. **On `main`:** Verify that sending reactions/comments fails to decrypt for messages sent via WhatsApp Web in standard 1-on-1 chats (`@s.whatsapp.net`) due to the hardcoded LID. (Mobile is using `@lid`, WA Web is using `@s.whatsapp.net`, haven't got to why yet)
2. **On `fix/poll_vote`:** Verify that reactions and comments now successfully work across *all* chat types (`@s.whatsapp.net`, `@g.us`, and `@lid`).

If no one gets to it, I'll test these remaining functions whenever I have some free time!

PR #1100, opened 2026-03-23
feat: exported error/confirmation receipt handling for manual sync

gusquadri:
Add the capability to send server-error receipts manually (not supported before) and export the confirmation receipt to be able to send it manually too later with the DisableManualHistorySyncReceipt true (new flag added to disable automatic confirmation receipts in manual sync mode)

### DIFF
--- a/internals.go
+++ b/internals.go
@@ -392,7 +392,7 @@ func (int *DangerousInternalClient) HandleDecryptedMessage(ctx context.Context, 
 }
 
 func (int *DangerousInternalClient) SendProtocolMessageReceipt(ctx context.Context, id types.MessageID, msgType types.ReceiptType) {
-	int.c.sendProtocolMessageReceipt(ctx, id, msgType)
+	_ = int.c.SendProtocolMessageReceipt(ctx, id, msgType)
 }
 
 func (int *DangerousInternalClient) DecryptMsgSecret(ctx context.Context, msg *events.Message, useCase MsgSecretType, encrypted messageEncryptedSecret, origMsgKey *waCommon.MessageKey) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


gusquadri:
Note: mediaKey is the encryption key for the downloadable history-sync blob itself

tulir:
That doesn't seem like it'd be common enough to be worth handling. If more control over history sync processing is needed, you should just use `ManualHistorySyncDownload`. It might make sense to add a new flag to skip the receipt so the custom handler can call it later.

gusquadri:
@tulir so dont you think is useful to add support to server-error receipt sending? If you consider not useful I can just add the flag you mentioned and revert other changes

tulir:
Having that as an exported method might be fine, so anyone doing manual downloads can call it if necessary

tulir:
Oh and I guess there'll need to be an exported way to send the protocol message receipt as well

gusquadri:
@tulir okay, thanks for the feedback! I'll make the changes to keep the error receipt as an exported function and just add the flag to skip the confirmation receipt sending if the user wants, reverting the automatic path logic changes I've made

internals.go
```diff
@@ -392,7 +392,7 @@ func (int *DangerousInternalClient) HandleDecryptedMessage(ctx context.Context,
 }
 
 func (int *DangerousInternalClient) SendProtocolMessageReceipt(ctx context.Context, id types.MessageID, msgType types.ReceiptType) {
-	int.c.sendProtocolMessageReceipt(ctx, id, msgType)
+	_ = int.c.SendProtocolMessageReceipt(ctx, id, msgType)
```
tulir: https://github.com/tulir/whatsmeow/blob/main/internals.go#L1 (just run `go generate`)

gusquadri: Oh, sorry for not reading this first line, I just regenerated it properly now

PR #1102, opened 2026-03-25
feat: initial cstoken support

gusquadri:
Introduce the initial implementation of cstokens, which currently appears to be in a very early stage in WA Web, as the feature flag is usually disabled in my tests.

It should receive the server-side nctsalt to generate cstokens (only through history sync and app sync events), and should only be generated when the tctoken is missing and the salt exists (as a fallback).

Initial question for Tulir:

- Should we introduce an in-memory cache for these salts, or should we just store them in the database?


purpshell:
> * Should we introduce an in-memory cache for these salts, or should we just store them in the database?

I think these salts need to be stored inside the Store since it seems like WA doesn't regularly keep them updated per session. They can be loaded into memory during Connection to reduce DB overhead though.

store/sqlstore/store.go
tulir, line 1: This has conflicts from the tctoken changes

store/sqlstore/upgrades/14-nct-salt.sql
```diff
@@ -0,0 +1,6 @@
+-- v14: Add NCT salt table for cstoken derivation
```
tulir:
```suggestion
-- v14 (compatible with v8+): Add NCT salt table for cstoken derivation
```

appstate.go
```diff
@@ -234,6 +234,25 @@ func (cli *Client) dispatchAppState(ctx context.Context, name appstate.WAPatchNa
 	}
 	logEvt.Msg("Received app state mutation")
 
+	if len(mutation.Index) > 0 &&
+		mutation.Index[0] == appstate.IndexNCTSaltSync &&
+		(mutation.Operation == waServerSync.SyncdMutation_SET || mutation.Operation == waServerSync.SyncdMutation_REMOVE) {
```
tulir: If it's not expected to have any other index keys, should check the exact length. Also, there are no other operations than SET and REMOVE

```suggestion
	if len(mutation.Index) == 1 && mutation.Index[0] == appstate.IndexNCTSaltSync {
```

appstate.go
```diff
@@ -234,6 +234,25 @@ func (cli *Client) dispatchAppState(ctx context.Context, name appstate.WAPatchNa
...
+		var err error
+		if mutation.Operation == waServerSync.SyncdMutation_SET {
+			if salt := mutation.Action.GetNctSaltSyncAction().GetSalt(); len(salt) > 0 {
+				err = cli.storeNCTSalt(ctx, salt)
+			} else {
+				err = cli.clearNCTSalt(ctx)
```
tulir: Do nulls actually appear in SET operations to make this if/else necessary?

gusquadri: > Do nulls actually appear in SET operations to make this if/else necessary?

Its a defensive behavior that I brought from waweb: 

```
var n = (t = e.value) == null || (t = t.nctSaltSyncAction) == null ? void 0 : t.salt;
if (n == null) ... this.malformedActionIndex();
```

tulir: actually just noticed that `storeNCTSalt` already has a fallback, so this if is completely redundant

cstoken.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (cli *Client) storeNCTSalt(ctx context.Context, salt []byte) error {
+	if cli.Store == nil || cli.Store.NCTSalt == nil {
+		return nil
+	}
+	if len(salt) == 0 {
+		return cli.Store.NCTSalt.DeleteNCTSalt(ctx)
+	}
+	return cli.Store.NCTSalt.PutNCTSalt(ctx, append([]byte(nil), salt...))
```
tulir:
```suggestion
	return cli.Store.NCTSalt.PutNCTSalt(ctx, salt)
```

appstate.go
```diff
@@ -234,6 +234,25 @@ func (cli *Client) dispatchAppState(ctx context.Context, name appstate.WAPatchNa
...
+			}
```
tulir:
```suggestion
			err = cli.storeNCTSalt(ctx, mutation.Action.GetNctSaltSyncAction().GetSalt())
```

appstate.go
gusquadri: > actually just noticed that `storeNCTSalt` already has a fallback, so this if is completely redundant

yeah, I just see it too, sorry

PR #1110, opened 2026-04-01
pair: make QR codes compliant 

purpshell:
QR Codes are no longer compliant since a long time ago. I only found this out by randomly scanning a WhatsApp Web QR code using the native viewer and it actually functioned. 

This might be a detection vector to show the user warnings upon trying to log in, but there is no research regarding that at the moment..

You can find this updated logic that includes both the URL prefix and the suffix with the PLATFORM_ID of the device in the module WAWebLinkDeviceQrcode.react


JoaoArtur:
I tested here and the Codes are invalid.
IMG_1515

purpshell:
> I tested here and the Codes are invalid. IMG_1515

This has been fixed

PR #1119, opened 2026-04-21
receipt: add configurable worker pool to limit concurrent retry processing

zhamghaoran:
## Problem

When sending messages to large groups, a burst of receipt retry messages
arrives simultaneously. Each retry triggers `handleRetryReceipt`, which
does database reads/writes (session store, pre-keys, etc.). With unlimited
concurrency this causes a sudden spike in database pressure that can
degrade overall performance.

## Solution

Add a `ReceiptWorkerCount` field to `Client` that limits the number of
goroutines processing receipts concurrently:

- **0 (default)**: original behavior — each receipt spawns its own goroutine (unlimited concurrency, no breaking change)
- **N > 0**: a fixed pool of N workers drains a buffered channel (512), capping concurrent database operations

```go
cli := whatsmeow.NewClient(deviceStore, nil)
cli.ReceiptWorkerCount = 10 // set before Connect
cli.Connect()
```

Also refactors `handleReceipt` into `processReceiptTask` with a top-level panic recovery, and switches retry handling to `BackgroundEventCtx` so it isn't bound to the caller's request context lifetime.

## Impact

- No breaking change — default behavior is identical to before
- Operators dealing with large-group receipt storms can tune `ReceiptWorkerCount` to match their database capacity

### DIFF
--- a/client.go
+++ b/client.go
@@ -123,6 +123,12 @@ type Client struct {
 	incomingRetryRequestCounter     map[incomingRetryKey]int
 	incomingRetryRequestCounterLock sync.Mutex
 
+	receiptQueue chan receiptTask
+	// ReceiptWorkerCount controls the number of concurrent receipt workers.
+	// 0 (default) means unlimited concurrency (each receipt gets its own goroutine).
+	// Set before calling Connect.
+	ReceiptWorkerCount int
+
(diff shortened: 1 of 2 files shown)


tulir:
This changes the behavior of non-retry receipts and is generally unnecessarily complicated

PR #1137, opened 2026-05-09
user: switch UpdateBlocklist to use LIDs

zennn08:
### Checklist

* [x] I have read and followed the contributing guidelines at <https://docs.mau.fi/bridges/general/contributing.html>

### DIFF
--- a/user.go
+++ b/user.go
@@ -12,7 +12,9 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strconv"
 	"strings"
+	"time"
 
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -942,16 +944,52 @@ func (cli *Client) GetBlocklist(ctx context.Context) (*types.Blocklist, error) {
...
+				return nil, fmt.Errorf("failed to resolve PN for %s: %w", jid, err)
+			}
+			pnJID = pn
+		}
+	}
+
+	itemAttrs := waBinary.Attrs{
+		"jid":    lidJID,
+		"action": string(action),
+		"dhash":  strconv.FormatInt(time.Now().UnixMilli(), 10),
```
tulir: This is wrong, the dhash is supposed to be the the value from the cached blocklist (if any), but easiest to omit it since it's not required

```suggestion
```

Issue #1130, opened 2026-05-03
Original message displayed edited one not


Sergey842248:
If I recieve a message which was edited I only see the original one and the edited one won't be shown. My setup worked perfectly but since one or two days I experience this issue. Since the official Beeper app (which also relies on the matrix protocol) has the issue for me, I guess the issue seems to be on the site of whatsmeow. Maybe Whatsapp changed it's API or similar.

Logs from mautrix (I redacted numbers and domains):
```
docker logs -f --since=0s matrix-whatsapp-bridge-1
2026-05-03T14:59:39.919Z DBG <message from="136150496886855@lid" id="AC9800FFFC4C8B1E6CB27FB31306A941" notify="Tine" sender_pn="49...@s.whatsapp.net" t="1777820379" type="text"><enc type="msg" v="2"><!-- 130 bytes --></enc><reporting><reporting_tag>0112f1d12f0711d389a862bd04804e29070f9db1</reporting_tag><reporting_token v="2">f71d48886984c09dae8f857ce8d9d9c6</reporting_token></reporting></message> component=whatsmeow login_id=49... sublogger=Recv user_mxid=@parsa:matrix.mydomain.com
2026-05-03T14:59:39.919Z DBG Decrypting message from 136150496886855@lid component=whatsmeow login_id=49... user_mxid=@parsa:matrix.mydomain.com
2026-05-03T14:59:39.93Z DBG Stored message secret key for AC9800FFFC4C8B1E6CB27FB31306A941 component=whatsmeow login_id=49... user_mxid=@parsa:matrix.mydomain.com
2026-05-03T14:59:39.93Z DBG Forced LID DM sender to phone number in incoming message evt_type=message lid=136150496886855@lid login_id=49... message_id=AC9800FFFC4C8B1E6CB27FB31306A941 pn=49...@s.whatsapp.net user_mxid=@parsa:matrix.mydomain.com
2026-05-03T14:59:39.93Z DBG <receipt id="AC9800FFFC4C8B1E6CB27FB31306A941" to="136150496886855@lid" type="inactive"/> component=whatsmeow login_id=49... sublogger=Send user_mxid=@parsa:matrix.mydomain.com
[...]

PR #1138, opened 2026-05-09
Fix message decryption for edited messages

Sergey842248:
A fix for edited messages not arriving and the decryption error occurring if a message was edited.
Related to the issue #1130. 

### Checklist

* [x] I have read and followed the contributing guidelines at <https://docs.mau.fi/bridges/general/contributing.html>

P.S.: Thank you for all the effort you put into this project, I use it everyday!


purpshell (commented):
The actual pattern that WhatsApp uses must be uncovered rather than brute-forcing things. I'm not sure if whatsmeow has this patched yet but this same problem affects all message-secret based messageAddons (polls, enc reactions, announcement group replies etc.) as well..

PR #1151, opened 2026-05-23
events: expose historical poll-vote tallies embedded in HistorySync

vadika:
## Summary

`waWeb.WebMessageInfo.PollUpdates` (proto field 45) is the primary phone's
bundled record of prior `PollUpdateMessage` events for a poll. The field
is decoded into Go structs but nothing in whatsmeow reads it, so
companion devices that consume `events.HistorySync` see empty tallies on
every poll posted before the companion paired — even though the data is
right there in the blob.

The official WhatsApp clients show those tallies because they walk the
field themselves. Every whatsmeow-based client (mautrix-whatsapp, wuzapi,
etc.) has the same blind spot.

## Change

- `events.HistoricalPollVote` struct (already-decrypted — the proto
  carries plaintext `PollVoteMessage`, no msgsecret derivation needed)
- `(*events.HistorySync).HistoricalPollUpdates()` helper flattens every
  embedded vote across the blob into a single slice
- Doc comment on `HistorySync` pointing at the previously-undocumented
  field

Hashes match the shape `Client.DecryptPollVote` yields for live votes so
a consumer's tallying code path can be uniform across live and historical
sources.

Non-breaking — existing consumers ignore the new helper.

## Background

Discovered while building [yawac](https://github.com/vadika/yawac), a
SwiftUI macOS WhatsApp companion. Patch verified end-to-end there
(implementation initially lived in the consumer; upstreaming so other
consumers don't need to duplicate the walk).

## Test plan

- [x] `go build ./types/events/`
- [x] Verified against a real account: paired companion with multiple
      group polls that had pre-pairing votes; after the initial
      `HistorySync` arrived, `HistoricalPollUpdates()` returned the
      correct per-voter records with hashes matching the poll options'
      SHA-256(name) digests.

### DIFF
--- a/types/events/events.go
+++ b/types/events/events.go
@@ -242,8 +242,37 @@ type StreamError struct {
 type Disconnected struct{}
 
 // HistorySync is emitted when the phone has sent a blob of historical messages.
+//
+// Each [waE2E.WebMessageInfo] inside Data.Conversations[].Messages may carry
+// previously-bundled poll vote records on its PollUpdates field. Those
+// records are already decrypted (the proto field is the plaintext
+// PollVoteMessage, not the encrypted PollUpdateMessage wrapper) — the
+// primary phone assembles them when packaging the history sync so newly
+// paired companions can render existing poll tallies that they never
+// received as live PollUpdateMessage events.
+//
+// Consumers that care about poll tallies should iterate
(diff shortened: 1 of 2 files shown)


tulir:
In general, whatsmeow won't unwrap fields just for the sake of having them in a different format. Also, flattening the votes is quite a specific abstraction which is unlikely to be generally useful.

PR #1167, opened 2026-06-11
send: don't fetch prekeys for own device on every message

jlucaso1:
Every DM send currently triggers a prekey fetch IQ for the sender's own device. `encryptMessageForDevices` builds the session prefetch list from all devices returned by usync, but the "skip self" check only happens later in the encrypt loop. Since there's never a Signal session with yourself, the own device always ends up in `retryDevices` and the fetched bundle is never used. Easy to spot with debug logging, every send is preceded by:

```xml
<iq to="s.whatsapp.net" type="get" xmlns="encrypt">
  <key><user jid="559980000001:51@s.whatsapp.net" reason="identity"/></key>
</iq>
```

This skips ownJID/ownLID when building the prefetch list, with the same condition the encrypt loop already uses. Same fix applied to the v3 path in sendfb.go.

WA Web never hits this because `getFanOutList` (WAWebDBDeviceListFanout) drops the own device before any session check or prekey fetch:

```javascript
var i = o("WAWebWidFactory").createDeviceWidFromDeviceListPk(e.id, t.id, t.isHosted);
o("WAWebUserPrefsMeUser").isMeDevice(i) || s.set(i.toString(), i);
```

### DIFF
--- a/send.go
+++ b/send.go
@@ -1285,6 +1285,11 @@ func (cli *Client) encryptMessageForDevices(
 	sessionAddressToJID := make(map[string]types.JID, len(allDevices))
 	sessionAddresses := make([]string, 0, len(allDevices))
 	for _, jid := range allDevices {
+		if dsmPlaintext != nil && (jid == ownJID || jid == ownLID) {
+			// There's never a session with ourselves; prefetching one would
+			// put it in retryDevices and fetch prekeys on every message.
+			continue
+		}
 		encryptionIdentity := jid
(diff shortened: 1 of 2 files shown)


tulir:
Small inaccuracies here*, but the underlying idea is correct, thanks for spotting it!

(*specifically: you can technically have a session with yourself, in which case there won't be any extra query on every message. Filtering on `dsmPlaintext != nil` isn't correct either, otherwise you'll get group messages encrypted for yourself)

Issue #858, opened 2025-07-01
App state syncs sometimes fail with LTHash mismatches


tulir:
> failed to decode app state &lt;type&gt; patches: failed to verify patch vX: mismatching LTHash

fixalllow:
Same problem! This problem is very common~

igortrinidad:
I'm having this issue too...

tulir:
For reference, the way to work around this is to handle `*events.AppStateSyncError` and:

1. Try a full sync via `FetchAppState` if it hasn't been tried yet
2. If a full sync was already tried and the event is emitted again, try sending a recovery request (`BuildAppStateRecoveryRequest` + `SendPeerMessage`)
3. Finally, if the phone sends a delivery receipt for the app state recovery request, but doesn't send a response (for some users the primary device refuses to respond like that), you can force it to reset app state using `BuildFatalAppStateExceptionNotification`. However, that will unlink all devices, so you need to relogin afterwards

Issue #651, opened 2024-08-29
Failed to sync app state after notification: failed to decode app state regular patches: failed to verify patch v564: mismatching LTHash


brunoaduarte:
@tulir , I'm getting this error after I created a label via WhatsApp Web.

```
04:16:22.460 [Client/Recv DEBUG] <notification from="s.whatsapp.net" id="484950606" t="1724915782" type="server_sync"><collection name="regular" version="588"/></notification>
04:16:22.466 [Client DEBUG] Got server sync notification that app state regular has updated to version 588
04:16:22.466 [Client/Send DEBUG] <ack class="notification" id="484950606" to="s.whatsapp.net" type="server_sync"/>
04:16:22.467 [Client/Send DEBUG] <iq id="103.122-15" to="s.whatsapp.net" type="set" xmlns="w:sync:app:state"><sync><collection name="regular" return_snapshot="false" version="563"/></sync></iq>
04:16:22.828 [Client/Recv DEBUG] <iq from="s.whatsapp.net" id="103.122-15" type="result">
  <sync>
    <collection name="regular">
      <patches>
        <patch><!-- 255 bytes --></patch>
        <patch><!-- 443 bytes --></patch>
        <patch><!-- 255 bytes --></patch>
        <patch><!-- 901 bytes --></patch>
        <patch><!-- 613 bytes --></patch>
        <patch><!-- 697 bytes --></patch>
        <patch><!-- 713 bytes --></patch>
        <patch><!-- 597 bytes --></patch>
        <patch><!-- 647 bytes --></patch>
        <patch><!-- 1225 bytes --></patch>
        <patch><!-- 613 bytes --></patch>
        <patch><!-- 917 bytes --></patch>
        <patch><!-- 1087 bytes --></patch>
        <patch><!-- 663 bytes --></patch>
        <patch><!-- 647 bytes --></patch>
        <patch><!-- 613 bytes --></patch>
        <patch><!-- 713 bytes --></patch>
        <patch><!-- 647 bytes --></patch>
        <patch><!-- 647 bytes --></patch>
        <patch><!-- 1053 bytes --></patch>
        <patch><!-- 663 bytes --></patch>
        <patch><!-- 867 bytes --></patch>
        <patch><!-- 679 bytes --></patch>
        <patch><!-- 1127 bytes --></patch>
        <patch><!-- 581 bytes --></patch>
      </patches>
    </collection>
  </sync>
</iq>
[...]

PR #1171, opened 2026-06-11
appstate: add SkipBrokenAppStatePatches opt-in for graceful recovery from unverifiable patches

### LINKED ISSUE #382: 409 conflict error
I always get an error when I mark mute or archive
`Error changing chat's mute state: server returned error updating app state: <collection name="regular_low" type="error"><error code="409" text="conflict"/><patches><patch><!-- 830 bytes --></patch><patch><!-- 700 bytes --></patch><patch><!-- 732 bytes --></patch><patch><!-- 750 bytes --></patch><patch><!-- 700 bytes --></patch><patch><!-- 1000 bytes --></patch><patch><!-- 660 bytes --></patch><patch><!-- 846 bytes --></patch><patch><!-- 774 bytes --></patch><patch><!-- 732 bytes --></patch><patch><!-- 1250 bytes --></patch><patch><!-- 1302 bytes --></patch><patch><!-- 894 bytes --></patch><patch><!-- 756 bytes --></patch><patch><!-- 628 bytes --></patch><patch><!-- 700 bytes --></patch><patch><!-- 732 bytes --></patch><patch><!-- 910 bytes --></patch><patch><!-- 756 bytes --></patch><patch><!-- 926 bytes --></patch><patch><!-- 854 bytes --></patch><patch><!-- 1000 bytes --></patch><patch><!-- 1222 bytes --></patch><patch><!-- 732 bytes --></patch><patch><!-- 814 bytes --></patch><patch><!-- 870 bytes --></patch></patches></collection>`

version: 2.2319.9

auroracapital:
## Problem

Several linked issues (#382, #518, #651, #858) describe a failure mode where one or more patches in an app state collection (most commonly `regular_low`, which carries archive/mute/pin state) permanently fail LTHash verification:

```
failed to decode app state regular_low patches: failed to verify patch v<N>: mismatching LTHash
```

or reference a sync key that the linked device never received:

```
failed to decode app state regular_low patches: failed to get key <X> to decode mutation: didn't find app state key
```

Because `DecodePatches` returns an error on the first bad patch, `fetchAppState` aborts the entire sync at that version. The consequences chain:

1. The stored version cursor stays at or below the bad patch — so every subsequent incremental notification triggers the same abort.
2. `SendAppState` (archive/mute/pin writes via `BuildArchive` etc.) encodes patches against the stale local version, and the server rejects them with a 409 conflict because the client is behind.
3. The 409 retry path calls `applyAppStatePatches` on the conflicting patches — which hits the same bad patch and also aborts — so the write never succeeds.

The bad patches originate server-side, typically written by another linked device. They repeat on every sync. The only documented user workaround (issue #518) is unlinking all devices.

## Root cause

`DecodePatches` and `validatePatch` (via `validateSnapshotMAC`) treat `ErrMismatchingLTHash` and `ErrKeyNotFound` as hard failures that abort the whole patch list. There is no mechanism to skip past a patch that is permanently unverifiable.

## Fix

This PR adds an opt-in field to `Client`:

```go
// SkipBrokenAppStatePatches can be set to true to enable graceful handling of app state
// patches that permanently fail LTHash verification or reference missing sync keys.
// When enabled, FetchAppState and SendAppState will advance the stored version cursor past
// any unverifiable patch and retry fetching from the next version, rather than aborting the
// entire sync. Skipped patches are lost (their mutations are not applied), which is the same
// recovery behaviour used by WhatsApp Web itself when it requests a snapshot reset.
//
// This is opt-in because in the normal case a verification failure indicates a local bug or
// data corruption that should be surfaced, not silently skipped. Enable it only when you have
// a chain of server-side bad patches that the primary device cannot fix (see GitHub issues
// #382, #518, #651, #858).
SkipBrokenAppStatePatches bool
```

When `true`, `FetchAppState` uses a bounded skip-and-retry loop (`fetchAppStateWithSkip`, up to 200 iterations):

1. Call `fetchAppState` normally.
2. If the error is `ErrMismatchingLTHash` or `ErrKeyNotFound`, read the current stored version, write `version+1` with a zeroed hash via `PutAppStateVersion`, sleep 300 ms (to avoid exhausting the per-connection IQ budget when many patches need skipping), then retry from the new cursor.
3. Repeat until the fetch succeeds or the iteration limit is hit.

`SendAppState`'s 409-conflict retry path is also covered: if `applyAppStatePatches` fails with a skippable error and the flag is set, it falls back to `fetchAppStateWithSkip` before retrying the send.

**Default behaviour is unchanged.** With `SkipBrokenAppStatePatches == false` (the default), every verification failure still aborts and surfaces errors exactly as before.

## Trade-offs

Skipped patches are lost — their mutations (archive/mute/pin toggles written by another device at that version) are not applied locally. This is the same trade-off WhatsApp Web accepts when it requests a fatal-recovery snapshot: the snapshot is authoritative but does not contain the intermediate mutations. In practice, the affected collections contain UI preference state (archive/mute/pin); losing a few patches from a poisoned chain is far less disruptive than a permanently broken sync.

The 300 ms throttle and 200-iteration cap bound both the wall-clock time and the server-side IQ cost of the recovery.

## Production validation

We ran this fix (as a downstream patch to the bridge binary at [lharries/whatsapp-mcp](https://github.com/lharries/whatsapp-mcp)) against an account whose `regular_low` chain had accumulated ~340 unverifiable patches (mix of `ErrMismatchingLTHash` and `ErrKeyNotFound` after a manual sync-key wipe). Results on 2026-06-10:

- The skip loop cleared all 340 bad patches in a single `FetchAppState` call (~340 × 300 ms ≈ 102 s).
- After the loop completed, `GetAppStateVersion` returned a version at or above the server head.
- 31/31 subsequent `SendAppState(BuildArchive(...))` calls succeeded and archive state propagated to the primary phone within seconds.

## Files changed

- `client.go` — add `SkipBrokenAppStatePatches bool` field with godoc
- `appstate.go` — add `fetchAppStateWithSkip`, `isSkippableAppStateError`, `extractVersionFromSkipError`; route `FetchAppState` and the 409 retry in `sendAppState` through the skip loop when the flag is set

`go build ./...`, `go vet ./...`, and `go test ./...` all pass on the patched tree.

Fixes #382, #518, #651, #858

### DIFF
--- a/appstate.go
+++ b/appstate.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -28,7 +29,15 @@ import (
 
 // FetchAppState fetches updates to the given type of app state. If fullSync is true, the current
 // cached state will be removed and all app state patches will be re-fetched from the server.
+//
+// When Client.SkipBrokenAppStatePatches is true, any patch that permanently fails LTHash
+// verification or references a missing sync key is skipped: its version is written to the store
+// with a zeroed hash so the next fetch starts from the following version. See that field's
+// documentation for a full description of the trade-offs.
 func (cli *Client) FetchAppState(ctx context.Context, name appstate.WAPatchName, fullSync, onlyIfNotSynced bool) error {
+	if cli.SkipBrokenAppStatePatches {
+		return cli.fetchAppStateWithSkip(ctx, name, fullSync, onlyIfNotSynced)
+	}
 	eventsToDispatch, err := cli.fetchAppState(ctx, name, fullSync, onlyIfNotSynced)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


tulir:
This is incorrect: in general, the LTHash covers the entire collection of app state patches, skipping any of them will propagate the error and the hash will be invalid forever. There are some rare cases where skipping it will work, but I don't know of any reliable way to tell the difference between that and a patch that's supposed to work but doesn't. There's no point in building in complicated hacks that only work in a small fraction of cases.

The correct solutions is sending an app state recovery request, though for some users the primary device refuses to respond to those. When the phone refuses (= sends a receipt but no response) the only option is sending a fatal app state exception notification, which will unlink all devices and merge the entire app state into a single patch.

PR #1186, opened 2026-06-30
pair: add support for passkeys

### LINKED ISSUE #1184: Passkey support
WhatsApp has started requiring some users to use passkeys when linking new devices

tulir:
Fixes #1184

### DIFF
--- a/client.go
+++ b/client.go
@@ -186,7 +186,10 @@ type Client struct {
 
 	BackgroundEventCtx context.Context
 
-	phoneLinkingCache *phoneLinkingCache
+	phoneLinkingCache    atomic.Pointer[phoneLinkingCache]
+	passkeyLinkingCache  atomic.Pointer[passkeyLinkingCache]
+	passkeyHandoffKey    atomic.Pointer[passkeyHandoffKey]
+	passkeySkipHandoffUX atomic.Bool
 
 	uniqueID  string
 	idCounter atomic.Uint64
--- a/internals.go
+++ b/internals.go
@@ -199,6 +199,18 @@ func (int *DangerousInternalClient) HandleConnectSuccess(ctx context.Context, no
 	int.c.handleConnectSuccess(ctx, node)
 }
 
+func (int *DangerousInternalClient) GenerateCsToken(ctx context.Context, jid types.JID) []byte {
+	return int.c.generateCsToken(ctx, jid)
+}
+
+func (int *DangerousInternalClient) StoreNCTSalt(ctx context.Context, salt []byte) error {
+	return int.c.storeNCTSalt(ctx, salt)
+}
+
+func (int *DangerousInternalClient) ClearNCTSalt(ctx context.Context) error {
+	return int.c.clearNCTSalt(ctx)
+}
+
 func (int *DangerousInternalClient) DownloadAndDecrypt(ctx context.Context, url string, mediaKey []byte, appInfo MediaType, fileEncSHA256, fileSHA256 []byte) (data []byte, err error) {
 	return int.c.downloadAndDecrypt(ctx, url, mediaKey, appInfo, fileEncSHA256, fileSHA256)
 }
@@ -503,6 +515,26 @@ func (int *DangerousInternalClient) SendPairError(ctx context.Context, id string
 	int.c.sendPairError(ctx, id, code, text)
 }
 
+func (int *DangerousInternalClient) HandlePasskeyNotification(ctx context.Context, node *waBinary.Node) {
+	int.c.handlePasskeyNotification(ctx, node)
+}
+
+func (int *DangerousInternalClient) TryHandlePasskeyContinuationNotification(ctx context.Context, node *waBinary.Node) {
+	int.c.tryHandlePasskeyContinuationNotification(ctx, node)
+}
+
+func (int *DangerousInternalClient) HandlePasskeyContinuationNotification(ctx context.Context, node *waBinary.Node) error {
+	return int.c.handlePasskeyContinuationNotification(ctx, node)
+}
+
+func (int *DangerousInternalClient) GetCompanionRef(ctx context.Context) (string, error) {
+	return int.c.getCompanionRef(ctx)
+}
+
+func (int *DangerousInternalClient) GetPasskeyRequestOptions(ctx context.Context) (*types.WebAuthnPublicKey, error) {
+	return int.c.getPasskeyRequestOptions(ctx)
+}
+
 func (int *DangerousInternalClient) GetServerPreKeyCount(ctx context.Context) (int, error) {
 	return int.c.getServerPreKeyCount(ctx)
 }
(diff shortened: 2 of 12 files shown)


VitorS0uza (approved):
approve

marcelotadeujr (approved):

rafacpti23:
Happy to be working with you. Congratulations on the great work!

jesselopespereira:
Thank you guys for this PR.

jucenir-code:
aprove

Alexandre-Gavazzi:
## ✅ Confirmed workaround for accounts stuck in per-connection passkey mode

If your number is in the aggressive Shortcake mode (428 on every connection attempt, even after using the browser-auth bridge from @familymachlin-git / #2676), here's what worked for us today:

1. **Delete your WhatsApp Business account entirely** — Settings → Account → Delete my account.
2. **Register the same number in WhatsApp Messenger** (the regular app, not Business).
3. **DO NOT** reserve a username / link Instagram / link Facebook Business (these seem to be triggers of the aggressive rollout).

That's it. Pure Baileys 7.0.0-rc.13 with `Browsers.appropriate('Chrome')` → fresh QR → 515 (restart required, normal) → `open` → sending messages. **No bridge needed.**

**Downsides:** you lose groups, chat history, media. Groups don't restore even with backup (admins need to re-add you).

**Theory:** the per-connection passkey rollout seems to target Business accounts and accounts linked to Meta Business Suite. A fresh Personal registration on the same number escapes it — at least for now.

Tested today on a `+55XXX` number that had been in the 428 loop for 24h+. Same number, same SIM, just deleted the Business account and re-registered as Personal on WhatsApp Messenger.

Everything Diego (@diegofelipeee), @familymachlin-git, @gobeyondpty, @willianclic and @marcoscarraro shared here is gold and still works for accounts where you can complete the browser login. But if your browser itself refuses to complete login (Business accounts often can't), **delete + recreate as Personal** is the emergency exit.

---

### ⚠️ Before you nuke your account — MAKE A GROUP LIST

You **will lose every group**. Even the ones you administer. Even with backup restored. Admins need to re-add you one by one. **Do this before deleting:**

1. Open every group you care about → screenshot the group info (name + admin numbers).
2. If you're the sole admin of any group → promote someone else first, otherwise the group dies with your account.
3. Save contacts of key admins in a separate note — you'll need to DM them from the new account to be re-added.
4. Same for media / documents you might need later — export what matters, WhatsApp backup won't cover this reliably.

---

### 🪦 The graveyard of things we tried that FAILED miserably (~2 full days of pain)

Posting this so nobody else wastes time going down the same rabbit holes:

1. **Baileys pure QR** → 428 instantly.
2. **Business → switch to WhatsApp Messenger (WITHOUT deleting)** → still asks for passkey. The switch alone doesn't fix it, you have to nuke the account.
3. **Browser ID spoofing** — tried `Browsers.macOS('Desktop')`, `Browsers.windows('Desktop')`, `Browsers.ubuntu('Chrome')`, `Browsers.macOS('Safari')`, default `appropriate('Chrome')` → all 428.
4. **Pairing code (8 digits)** instead of QR → server closes the socket before you can even type it.
5. **whatsapp-web.js** (Puppeteer) → the modal "Quick security check with Passkey" returns "Something went wrong" forever.
6. **Venom-bot** → same thing, Puppeteer-based, same block.
7. **WPPConnect** → same, ditto.
8. **`webVersionCache: remote`** trick (@mehul-vekariya's fix) → unblocked it for some, for us reached the passkey modal but hit the same wall.
9. **`puppeteer-extra-plugin-stealth`** → removed the "controlled by automation" bar but the passkey check still failed.
10. **Virtual WebAuthn authenticator via CDP** (`WebAuthn.addVirtualAuthenticator`) → `NotAllowedError` — server rejects fake credentials (`.get()` with `allowCredentials: []` requires a discoverable credential we don't have).
11. **caBLE / Hybrid Transport (cross-device passkey via Bluetooth)** → requires Bluetooth LE on the PC. No BT adapter = dead end. Even with BT, it's a manual click per connection — not viable in production.
12. **Copy Chrome profile (7.8 GB!)** to Puppeteer `userDataDir` → passkeys live in Google Password Manager and require active Sync — Sync doesn't survive the copy.
13. **Puppeteer connect via `--remote-debugging-port=9222`** with the real Chrome → Chrome ignored the flag because another process was alive, port never opened.
14. **Browser-auth bridge (@familymachlin-git technique)** via Console script → creds extraction WORKED (Curve.verify OK, companion device correctly derived, 16 Noise candidates generated) → Baileys reached **`logging in...`** instead of `attempting registration` (progress!) → but server still killed the socket with 428 after ~200ms. Confirmed the aggressive rollout blocks even valid re-logins for Business accounts.
15. **Cycling through Noise candidates [0..7]** → all 428.
16. **Fake network fingerprints, User-Agent shenanigans, hardcoded Web version pinning** → nothing.
17. **Waited for a maintainer response on the mainline fork/PR** → all still in `REVIEW_REQUIRED` limbo.
18. **Considered forking Baileys ourselves to implement Shortcake natively** → mathematically impossible headless (WebAuthn requires a user gesture + real authenticator, that's the whole point of the spec).

At some point I was one Bluetooth USB dongle away from ordering it on Amazon just to test caBLE. Then I remembered @JoshAlfaro's comment saying only Business accounts were affected — deleted the account, recreated as Personal, connected on the first QR. **20 seconds. That easy.**

Two days of engineering warfare killed by a 30-second setting change. 🪦 R.I.P. browser-auth bridge, we hardly knew ye.

— @Alexandre-Gavazzi

VitorS0uza:
Now, we need a way to handle the generation of the WebAuthnResponse.

mircoianese:
I'm not sure this PR is enough to solve the passkey authentication issue, unfortunately. As I see it, the only way for this to work is if the whatsmeow instance has the passkey stored. So we'll probably need to add passkey generation somehow.

tulir:
This PR is enough, acquiring the passkey response is left as an implementation detail. There are plenty of solutions for that, but none are in scope for whatsmeow. For reference, the flow is:

1. Pair normally
2. Wait for `*events.PairPasskeyRequest`
3. Use the PublicKey field inside the event to get a response from the passkey, e.g. with [`navigator.credentials.get()`](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/get). The JSON is compatible with [`PublicKeyCredential.parseRequestOptionsFromJSON()`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/parseRequestOptionsFromJSON_static)
4. Call `client.SendPasskeyResponse()` with the result JSON ([`PublicKeyCredential.toJSON()`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/toJSON) produces compatible output)
5. Wait for `*events.PairPasskeyConfirmation`
6. This should never happen currently, but if `SkipHandoffUX` is `false`, I think the user is supposed to confirm that the phone is showing the same code as what the event contains (I assume this is for a passkey-only pairing that doesn't need the old qr or phone code step)
7. Call `client.SendPasskeyConfirmation()`

If using GetQRChannel, it automatically does step 7 when SkipHandoffUX is false, so only the `QRChannelEventPasskeyRequest` event needs to be handled manually

PR #1201, opened 2026-07-09
call: add first-class 1:1 call signaling and media-handoff API

purpshell:
Adds a stateful 1:1 call control plane to whatsmeow while leaving RTP/SRTP and codecs to external media implementations.

Includes:
- outgoing and incoming call lifecycle APIs
- call-key encryption and decryption
- relay parsing and direction-aware endpoint election (FNA for incoming, authenticated non-FNA for outgoing)
- typed media-ready/media-stop, mute, ack, and video events
- initial video negotiation and independent mid-call video transitions
- typed video acknowledgements with companion routing preserved
- answering-device identity updates for media rekeying

The branch is rebased onto current main and is consumed by purpshell/meowcaller#8 as the media backend boundary.

Validation:
- go test ./...
- go vet ./...
- go test -race .


types/call_test.go
purpshell, line 1: I didn't know what to do with the test files when porting over so i just kept them. I'm also not sure of the directory / file naming structure, maybe a voip/ dir would be good.

andresoodeen:
We are seeing the same current WhatsApp relay failure on a QR-linked companion implementation using the latest Meowcaller revision: outbound audio is initially accepted by the remote device, but inbound media reaches us only as a small burst of MLow/SID packets before it stops. The relay allocation and consent pings continue successfully, so this does not look like an ICE/STUN liveness failure.

We send audio from the start and tested immediate/deferred accept handling, capability and voip-settings variants, and multiple relay candidates. None restored sustained inbound speech. No account details, relay credentials, or packet captures are being shared here.

Does this 1:1 call-control/media-handoff PR include the current post-accept relay transition or signalling needed to keep inbound media flowing? If not, is there a known branch, required accept field, mute/keepalive sequence, or reproducible current-client trace that addresses it? We can validate a safe patch and report only redacted results.

mnofresno:
What's missing so that this PR gets merged? let me know if I can help you with anything.

andresoodeen:
Hello,

You have received a newsletter from Blue Chip Technologies Ltd..

However, your email software can't display HTML emails. You can view the newsletter by clicking here:

{link-id:190893449466086826}



You're receiving this newsletter because you have shown interest in Blue Chip Technologies Ltd..
Not interested anymore? Click here to unsubscribe:
{link-id:190893449460843945}

andresoodeen:
Hi Mariano,

Thanks for offering to help. We are using the MeowCaller/whatsmeow 1:1 call
path as the WhatsApp side of an Asterisk AudioSocket bridge, and we have
been chasing a persistent direction-specific media failure.

What we consistently observe:
- Signaling completes, relay allocation and consent succeed, WARP
authentication succeeds, and the gateway continuously sends valid 60 ms
MLow frames (16 kHz mono with 960-sample RTP timestamp increments).
- Asterisk/3CX to WhatsApp is usually audible.
- WhatsApp mobile app to the gateway often delivers no RTP/RTCP after
accept, or only a brief burst of silent/SID frames. We see no SRTP
unprotect failures.
- WhatsApp then shows "reconnecting" and the call eventually ends.
- WhatsApp Desktop has produced healthy bidirectional calls on the same
bridge. Direct SIP echo, Asterisk AudioSocket, and PBX-side tests are
clean, so the failure appears to be in the mobile-app 1:1
relay/media-handoff path before Asterisk.

We have already tested:
- ordinary versus participant-aware Allocate;
- FNA versus non-FNA and each advertised relay independently;
- continuous encoded silence from call start;
- MLow framing/timing, stable SSRC, and sequence/timestamp continuity;
- immediate versus deferred accept, capability/voip_settings variants, and
offer metadata;
- mute_v2 handling and removal of speculative automatic mute/unmute;
- type-1/type-9 transport transitions and relaylatency;
- legacy subscription bindings enabled and disabled;
- a second media engine, which showed the same zero-inbound-RTP behavior;
- lower logging pressure plus AudioSocket/DSP checks; healthy traces show
zero local drops.

One newer finding concerns multi-device calls. A linked Desktop can send
preaccept while it is only ringing. If preaccept is treated as the final
answer, the mobile may receive accepted_elsewhere. We have therefore
separated provisional media setup from formal answer: preaccept may prepare
the type-1/net-only relay transition, while only formal accept selects and
rekeys the answering device and dismisses siblings. One outbound trace then
produced hundreds of non-silent decoded return frames, but the result is
not yet consistently reproducible.

Could you or purpshell help clarify:
1. The exact current mobile-app preaccept -> relaylatency ->
transport/mute_v2 -> accept sequence expected by PR #1201.
2. Whether the relay token/type-1 transition must be sent to every
preaccepting device or only to the eventual formal accepter.
3. Which identity should drive receive-key derivation/rekeying when
preaccept comes from a companion LID but formal accept comes from the
primary or bare LID.
4. Whether there is a sanitized current Android/iOS 1:1 trace or test
fixture that shows sustained inbound media after accept.

We can provide redacted per-call signaling timelines and RTP counters, and
we can test a focused patch quickly. If there is a particular missing test
or API edge preventing #1201 from being merged, we would also be happy to
validate it against our Asterisk bridge and report the results.

Best regards,
Andre

On Mon, 10 Aug 2026 at 19:22, Mariano Martín Fresno <

ochenane:
Hi, is there any updates on this?
Are there any way we can help make this happen faster?

andresoodeen:
I switched to https://github.com/oxidezap/whatsapp-rust and this has been
working since.

PR #1211, opened 2026-07-30
feat(logging): dump raw stanza on 401 device_removed

Prodigy90:
Diagnostic for a live WASBOT issue: ~35% of successfully-linked sessions are torn down by `401 device_removed`, many within seconds of linking, and none recover.

The existing handler extracts only `code` and the `conflict` type, discarding the rest of the stream error node. If the server sets any additional attribute (a reason, an actor, a timestamp) it is currently invisible — and it is the only protocol-level lead on whether the removal is user-initiated or server-side anti-abuse.

Logs the full node at warn level so it survives the production log filter.

No behavior change.


Prodigy90:
Opened against the wrong remote — this is a private-fork diagnostic, not an upstream proposal. Closing.

PR #1234, opened 2026-08-26
fix: handle PairPhone passkey handoff notifications

### LINKED ISSUE #1233: PairPhone fails after passkey verification with missing link_code_pairing_wrapped_primary_ephemeral_pub
### Environment

- whatsmeow: `v0.0.0-20260821141805-33cfac511629`
- Go: `1.26.0`
- Platform: Linux server
- WhatsApp client: [Android/iOS, version if known]
- Account type: [WhatsApp Personal / WhatsApp Business]
- Pairing method: `Client.PairPhone`
- Client type: `whatsmeow.PairClientChrome`
- Client display name: `Chrome (Linux)`

### Problem

Phone-number pairing intermittently fails after WhatsApp accepts the pairing code.

The server successfully generates and displays a pairing code. The user enters the code in:

WhatsApp → Settings → Linked Devices → Link a Device → Link with Phone Number

After the code is submitted, WhatsApp appears to continue the linking process, but the pairing does not complete. whatsmeow logs:

```text
Failed to handle code pair notification:
missing <link_code_pairing_wrapped_primary_ephemeral_pub> element in notification
```

### Reproduction
1. Start a new whatsmeow client with an empty device store.
2. Call Connect().
3. Wait for the initial events.QR event.
4. Call: 
```
pairingCode, err := client.PairPhone(
    ctx,
    phone,
    true,
    whatsmeow.PairClientChrome,
    "Chrome (Linux)",
)
```
5. Enter the generated code in WhatsApp.
6. Confirm the linking request on the phone.
7. If prompted, complete passkey/biometric/PIN verification.
8. Observe that pairing fails and the following notification error is logged: `missing <link_code_pairing_wrapped_primary_ephemeral_pub> element in notification`

### Observed behavior
The failure is intermittent from the application perspective. Several attempts may fail after generating a valid code. In one case, after the same phone number eventually paired successfully, removing the account from the application interface and pairing it again did not reproduce the issue.

The failure occurs after code generation, rather than while generating the code.

Relevant log excerpts:
```
[api] phone session <session-id>: pairing code issued
Failed to handle code pair notification:
missing <link_code_pairing_wrapped_primary_ephemeral_pub> element in notification
```
The error has occurred for multiple fresh pairing sessions.

### What we tried
- Waiting for the initial events.QR event before calling PairPhone.
- Creating a fresh whatsmeow client for every attempt.
- Using a fresh pre-pairing device state.
- Retrying the pairing flow.
- Disconnecting failed clients promptly.
- Trying again after a previous successful pairing.
- Updating to the current whatsmeow pseudo-version listed above.
The timing change did not eliminate the notification error.

### Questions
1. Does this notification correspond to the new passkey-required linking flow introduced by WhatsApp/Meta?
2. Is PairPasskeyRequest expected to be emitted before this notification?
3. Can phone-code pairing currently be completed when WhatsApp requires passkey verification?
4. Is the missing link_code_pairing_wrapped_primary_ephemeral_pub element caused by an unsupported notification variant, or is the client expected to retain additional state from the passkey request?
5. Is there a recommended workaround, such as QR pairing, a specific client configuration, or a newer commit?
6. Could whatsmeow ignore this notification error and continue, or does the missing element prevent the cryptographic pairing response from being generated?

### Additional information
The problem appears to be related to the account's first successful companion-device pairing.

Before an account has ever paired successfully, repeated phone-code pairing attempts can generate valid codes but fail later with the missing-element notification.

After one attempt eventually pairs successfully, removing the account from the application's interface and repeating the login for that same number does not reproduce the original pairing failure.

Removing the account from the application releases the application enrollment and logs out the connected companion device. It does not reuse the previous temporary pre-pairing identity; each new attempt still calls Container.NewDevice().

This suggests that the successful pairing changes server-side WhatsApp account/linking state, rather than the application merely recovering a stale local device record.

### Possibly related issues
#1188  seems to report a the same error, but for Business accounts only. In my case the accounts are not Business.

mvanhorn:
Update `handleCodePairNotification` in `pair-code.go` to distinguish a notification for the active pairing reference that lacks the legacy wrapped-primary key from a complete legacy code-pair challenge. Treat the matched-reference, missing-key form as a passkey handoff precursor: do not attempt the legacy key-bundle response, and leave the connection ready for the existing passkey notification handler to emit `PairPasskeyRequest`. Continue rejecting notifications with no pending phone pairing or a mismatched reference, and preserve all length, identity-key, and cryptographic validation when the wrapped-primary key is present.

`Client.PairPhone` can receive a `link_code_companion_reg` notification after the phone accepts the code and performs passkey or biometric verification, but that notification may omit `link_code_pairing_wrapped_primary_ephemeral_pub`. The current `handleCodePairNotification` path assumes every matching registration notification is the legacy code-pair cryptographic challenge and logs an element-missing failure. The separate `passkey_prologue_request` route already exists, so this premature failure should not prevent or obscure the handoff into the passkey flow. The reported case is intermittent and most visible for an account's first successful companion-device pairing.

A notification for the pending pairing reference without `link_code_pairing_wrapped_primary_ephemeral_pub` is accepted as a handoff precursor and does not attempt the legacy `companion_finish` path.
A missing-key notification with no pending phone pairing still returns the existing pending-pairing error.
A missing-key notification whose pairing reference differs from the cached reference still returns the pairing-reference mismatch error rather than being mistaken for a handoff.
A legacy notification that includes a wrapped-primary key shorter than 80 bytes still returns the existing length error, proving the handoff handling does not weaken malformed-challenge validation.

### Checklist

* [ ] I have read and followed the contributing guidelines at <https://docs.mau.fi/bridges/general/contributing.html>

Fixes #1233

### DIFF
--- a/pair-code.go
+++ b/pair-code.go
@@ -166,7 +166,11 @@ func (cli *Client) handleCodePairNotification(ctx context.Context, parentNode *w
 	if string(linkCodePairingRef) != linkCache.pairingRef {
 		return fmt.Errorf("pairing ref mismatch in code pair notification")
 	}
-	wrappedPrimaryEphemeralPub, ok := node.GetChildByTag("link_code_pairing_wrapped_primary_ephemeral_pub").Content.([]byte)
+	wrappedPrimaryEphemeralPubNode, ok := node.GetOptionalChildByTag("link_code_pairing_wrapped_primary_ephemeral_pub")
+	if !ok {
+		return nil
+	}
+	wrappedPrimaryEphemeralPub, ok := wrappedPrimaryEphemeralPubNode.Content.([]byte)
 	if !ok {
 		return &ElementMissingError{
 			Tag: "link_code_pairing_wrapped_primary_ephemeral_pub",
--- a/pair-code_test.go
+++ b/pair-code_test.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2026 Tulir Asokan
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package whatsmeow
+
+import (
+	"context"
+	"testing"
+
+	waBinary "go.mau.fi/whatsmeow/binary"
+)
+
+func TestHandleCodePairNotificationPasskeyHandoff(t *testing.T) {
+	cli := &Client{}
+	cli.phoneLinkingCache.Store(&phoneLinkingCache{pairingRef: "pairing-ref"})
+	node := &waBinary.Node{Content: []waBinary.Node{{
+		Tag: "link_code_companion_reg",
+		Content: []waBinary.Node{{
+			Tag:     "link_code_pairing_ref",
+			Content: []byte("pairing-ref"),
+		}},
+	}}}
+
+	err := cli.handleCodePairNotification(context.Background(), node)
+	if err != nil {
+		t.Fatalf("expected passkey handoff precursor to be accepted, got %v", err)
+	}
+}
+
+func TestHandleCodePairNotificationMissingWrappedPrimaryKeyValidation(t *testing.T) {
+	tests := []struct {
+		name        string
+		cachedRef   string
+		notifiedRef string
+		wantError   string
+	}{
+		{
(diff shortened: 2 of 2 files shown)


DarkFighterLuke:
I tried your PR and I can confirm it works now! Thank you for your very quick help ❤️

PR #1257, opened 2026-09-05
binary: don't panic on malformed nodes

Kizuno18:
The node decoder had several unchecked type assertions and one unguarded slice expression that a single malformed frame could reach, and consumeFrames in socket/noisesocket.go has no recover(), so one bad frame from a buggy or hostile relay takes the whole process down. They now return ErrInvalidJIDType / ErrInvalidNode, and Unpack rejects empty input.
Based on [https://github.com/tulir/whatsmeow/pull/1160](https://github.com/tulir/whatsmeow/pull/1160) by @ThiagoBauken, plus:
- A truncated-input test covering frames cut off mid-token (short list header, short node tag, JID pair missing its server), which the original tests didn't exercise, and MPL headers on the new test files.
- A recursion depth cap. readNode and read recurse into each other unbounded, so a frame that nests a value inside itself for its whole length (a node whose only content is a list holding the next node, or a JIDPair whose user token is another JIDPair) goes as deep as it has bytes. A ~12.5 MB frame of either fits under socket.FrameMaxSize and blows the goroutine stack, which is fatal error: stack overflow, not a panic, so no recover() can catch it. The decoder now returns ErrNodeTooDeep past 256 levels; real stanzas are a handful of levels deep.
Happy to close this if @ThiagoBauken wants to add these to #1160.
Verified with go build, go vet, go test -race ./... and staticcheck.
- I have read and followed the contributing guidelines at [https://docs.mau.fi/bridges/general/contributing.html](https://docs.mau.fi/bridges/general/contributing.html)

### DIFF
--- a/binary/decoder.go
+++ b/binary/decoder.go
@@ -109,7 +109,7 @@ func (r *binaryDecoder) readPacked8(tag int) (string, error) {
 	}
 
 	ret := build.String()
-	if startByte>>7 != 0 {
+	if startByte>>7 != 0 && len(ret) > 0 {
 		ret = ret[:len(ret)-1]
 	}
 	return ret, nil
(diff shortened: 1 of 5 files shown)


jobasfernandes:
What a waste of artificial intelligence.

PR #1265, opened 2026-09-14
receipt: handle retry receipts synchronously when sync ack is enabled

Fizzadar:


tulir (approved):
Hmm, this is probably fine. Retry receipt handling does networking, but shouldn't be anything too blocking.

